Issue #4, opened 2016-03-12
Carrier phase is not calculated correctly


ljbade:
I captured a short RTCM3 log from the SYM10 Geoscience Australia station.

I then processed this with rtklib `convbin` to get a RINEX file.
I also processed with `rtcm32sbp`, `sbp2json`, and finally `sbp2rinex` to get a matching RINEX file after the RTCM3 to SBP conversion.

Analysing the two files shows that pseudorange is correctly converted, but that carrier phase is very different. Even more importantly the carrier phase delta between epochs does not match either.

I have attached the files.

RTCM log [sym10.rtcm3](https://files.zenhub.io/56e3734ea0b9fd8b56591c93)
SBP RINEX [sym10.sbp.obs](https://files.zenhub.io/56e3734fa0b9fd8b56591c94)
rtklib RINEX [sym10.obs](https://files.zenhub.io/56e3734fa0b9fd8b56591c95)
SBP [sym10.sbp.json](https://files.zenhub.io/56e3734fa0b9fd8b56591c96)

/cc @mfine @mookerji @denniszollo @fnoble

ljbade:
I have been comparing out RTCM impkementation with the spec.

So far some problems:
- we do not check for 80000h in DF011 GPS L1 Pseudorange, or 80000h DF012 GPS L1 PhaseRange – L1 Pseudorange. These values indicate the measurements are missing, and only used for computing L2. Thus we should probably drop this sat for this epoch.
- I don't think the carrier phase range computation is supposed to include DF014 GPS Integer L1 Pseudorange Modulus Ambiguity. This line https://github.com/swift-nav/gnss-converters/blob/master/haskell/src/Data/RTCM3/SBP.hs#L55 although the range of the RTKLIB cycles indicates that it does
- the way the integer and fractional parts of carrier phase are calculated, to not use same algorithim as in the firmware see https://github.com/swift-nav/gnss-converters/blob/master/haskell/src/Data/RTCM3/SBP.hs#L51-L53 vs https://github.com/swift-nav/piksi_firmware/blob/master/src/sbp_utils.c#L242-L251
- we don't compute the cycle slip indicator properly. In RTCM the value of DF013 GPS L1 Lock Time Indicator going backwards indicates a cycle slip. Hence the SBP lock value should be incremented

mfine:
The changes in https://github.com/swift-nav/gnss-converters/pull/16 can help with lock timer calculations.

ljbade:
[...]

PR #5, opened 2016-03-12
Fix carrier phase conversion

ljbade:
First step to fix all of https://github.com/swift-nav/gnss-converters/issues/4


ljbade:
@mfine Can you review?

PR #12, opened 2016-06-15
Fix up scaling

mfine:
Fix up the scaling factor - move it outside of the P calculation. Revert P in L calculation.

/cc @denniszollo @ljbade

### DIFF
--- a/haskell/src/Data/RTCM3/SBP.hs
+++ b/haskell/src/Data/RTCM3/SBP.hs
@@ -41,9 +41,9 @@ fromGpsObservationHeader hdr = do
     }
 
 toP :: GpsL1Observation -> GpsL1ExtObservation -> Word32
-toP l1 l1e = round $ p * 100 where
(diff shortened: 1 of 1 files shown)


ljbade:
Ah that looks better, I wonder how we can test it without need to deploy it to Skylark?

mfine:
If you have a local setup that uses `rtcm32sbp`, can run this branch locally to evaluate it.

ljbade:
hmm I'm currently using the published stack, how do I run a branch version?

mfine:
get this branch (bring in my remote branch), then run `stack install` in the `haskell` directory.

ljbade:
Is there a way to test this with the Piksi console also running? Or another way to both send the sbp and view the logging>

mfine:
@ljbade @denniszollo uses a TCP port and socat.

We can also just pull it in and try it out - you're the only active user

ljbade:
Yeah OK lets just pull it in

PR #14, opened 2016-06-29
Filter satellites with id's > 32

### LINKED ISSUE #9: Be careful of SBAS IDs
In RTCM any GPS IDs >32 are reserved, except for 40-58 which are used for SBAS.

We need to filter (and warn?) about any IDs outside these ranges, and assign the correct SBP system/constellation ID numbers to the SBAS IDs. Or drop those observations.

/cc @mfine

mfine:
Fixes #9.

Don't convert satellite id's > 32.

/cc @ljbade

### DIFF
--- a/haskell/src/Data/RTCM3/SBP.hs
+++ b/haskell/src/Data/RTCM3/SBP.hs
@@ -76,6 +76,7 @@ toSid sat = GnssSignal
 
 fromObservation1002 :: Observation1002 -> Maybe PackedObsContent
 fromObservation1002 obs =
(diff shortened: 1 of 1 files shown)


JoshuaGross:
@mark I think it's worthwhile adding comments about why we're skipping satellites >32 / links to relevant reading

ljbade:
Yeah, maybe ref the GPS ICD, and note that GPS PRNs are only 1-32, values over that are reserved for special uses like SBAS sats, which a CORS receiver might also be tracking.

ljbade:
Otherwise lgtm

PR #15, opened 2016-06-30
Sender id as Station id

### LINKED ISSUE #7: Track the station reference ID
It is possible for two stations' observations to be sent in one stream by using different Reference Station IDs.

We should check and latch the first ID we see. Then report an error/warning if we see a different ID, and drop those observations.

/cc @mfine

mfine:
Pull in the Station id as Sender id. Set the top byte.

Fixes #7.

/cc @ljbade


ljbade:
Excellent!

PR #20, opened 2016-07-21
1013 message MJD to WN

mfine:
Listen to 1013 messages and pick up MJD and use it to set WN.

/cc @mookerji @ljbade


ljbade:
Where does it handle the roll over?

mfine:
There's no rollover handling yet.

PR #23, opened 2016-07-28
Invalid measurements

mfine:
Filter out invalid measurements. DF011, DF012, and DF018 measurements with a value of `80000h` are invalid. Throw away observations if these values are set, along the following lines:

+ 1002: if DF011 or DF012 are invalid, throw away L1 obs
+ 1004: if DF011 or DF012 are invalid, throw away L1 and L2 obs; if DF018 is invalid, throw away L2 obs

The documentation indicates that the L2 obs can be processed while DF011 or DF012 are invalid; but we actively use those fields to produce a L2 obs.

/cc @mookerji @ljbade

### DIFF
--- a/haskell/src/Data/RTCM3/SBP.hs
+++ b/haskell/src/Data/RTCM3/SBP.hs
@@ -195,8 +195,8 @@ invalid_L1 l1 =
 -- | Determine whether an L1 + L2 RTCM observation is invalid.
 --
 -- See DF011, DF012, and DF018 of the RTCM3 spec
-invalid_L2 :: GpsL1Observation -> GpsL2Observation -> Bool
-invalid_L2 l1 l2 = invalid_L1 l1                     ||
+invalid_L2 :: GpsL2Observation -> Bool
+invalid_L2 l2 =
   l2 ^. gpsL2Observation_carrierMinusCode == 0x80000
 
 -- | Construct metric pseudorange (meters!) from L1 RTCM observation.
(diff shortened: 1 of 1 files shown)


mfine:
Let me know if this logic looks alright!

ljbade:
I think we also need to check DF017 for 2000h.

Also about L2, I think the spec says that if you get 8000h in the L1 values, you can still calculate L2 using the 8000h value as-is in the formula.

Should we do anything about 0 SNR value? I suppose SBP doesn't have a way to indicate SNR is missing.

haskell/src/Data/RTCM3/SBP.hs
```diff
@@ -184,6 +184,21 @@ mjdEpoch = 44244
...
+  l1 ^. gpsL1Observation_pseudorange == 0x80000      ||
+  l1 ^. gpsL1Observation_carrierMinusCode == 0x80000
+
+-- | Determine whether an L1 + L2 RTCM observation is invalid.
+--
+-- See DF011, DF012, and DF018 of the RTCM3 spec
+invalid_L2 :: GpsL2Observation -> Bool
+invalid_L2 l2 =
+  l2 ^. gpsL2Observation_carrierMinusCode == 0x80000
+
```
ljbade: Also check DF017

haskell/src/Data/RTCM3/SBP.hs
```diff
@@ -418,16 +433,17 @@ fromMsg1002 m = do
 fromObservation1004 :: MonadStore e m => Word16 -> Observation1004 -> m [PackedObsContent]
 fromObservation1004 station obs =
   -- Only lower set of PRN numbers (1-32) are supported
-  if sat > maxSats then return mempty else do
+  if sat > maxSats || invalid_L1 l1 then return mempty else do
     obs1 <- fromL1SatelliteObservation station sat l1 l1e
```
ljbade: It should be possible to skip obs1 when invalid_L1, but still output obs2?

haskell/src/Data/RTCM3/SBP.hs
mfine: Done.

haskell/src/Data/RTCM3/SBP.hs
mfine: Yeah, I don't follow what it's saying - I see a lot of the l1 values being used for the L2 calculations. What do you think the logic should look like here?

ljbade: Well I think what happens is, if the receiver can't track L1, but can do L2:
It sets the L1 values to 80000h, then adjusts the L2 values so that when using 80000h in the calculation it gets the correct value.

So basically if you detect bad L1, you dont send the observations, but you still use the result to calculate L2.

mfine: Ok, I'll update the logic to reflect that.

mfine: @ljbade check out e057c29b94f630c

ljbade:
Looks good to me

PR #34, opened 2017-02-21
Update to librtcm with Msg1019; decode Msg1019 to SBP MSG_EPHEMERIS_GPS.

JoshuaGross:
cc @ljbade @mfine @benjamin0


JoshuaGross:
need to add some tests

haskell/src/Data/RTCM3/SBP.hs
```diff
@@ -520,6 +534,52 @@ fromMsg1006 m =
     , _msgBasePosEcef_z = fromEcefVal $ m ^. msg1006_reference ^. antennaReference_ecef_z
     }
 
+-- | Convert an RTCM 1019 GPS ephemeris message into an SBP MsgEphemerisGps.
+fromMsg1019 :: MonadStore e m => Msg1019 -> m MsgEphemerisGps
```
mfine: Can this be broken up some similarly to other functions in this file? e.g., a function to generate `EphemerisCommonContent`, a function to generate `GpsTime`. Would help a bit with this big guy.

ljbade:
I think you need to check how the toe and toc need to be adjusted for week to over as the week number is not the full 10 bits. I think the firmware performs an adjustment after decoding the ephemeris but before it is saved

ljbade:
*rollover

mfine:
Some tests around the smaller functions would be helpful - this PR has added no tests.

ljbade:
looks ok, something worth testing would be to grab some ephemerides around the week rollover to test the week number logic

haskell/src/Data/RTCM3/SBP.hs
```diff
@@ -198,6 +206,27 @@ toGpsTime hdr = do
...
   liftIO $ modifyMap gpsTimeMap gpsTime station $ updateGpsTime tow
 
+-- | Produce GPS time from RTCM time (week-number % 1024, tow in seconds, scaled 2^4).
+-- Stateful but non-side-effecting.
+rtcmTimeToGpsTime :: MonadStore e m => Word16 -> Word16 -> m GpsTime
+rtcmTimeToGpsTime wn tow = do
+  stateWn <- view storeWn >>= liftIO . readIORef
+  let wn' = reconcileWn stateWn wn
+  return $ GpsTime
+    { _gpsTime_tow = 16 * (fromIntegral tow)
```
mfine: nit the parens should be unnecessary - just `16 * fromIntegral tow`

haskell/src/Data/RTCM3/SBP.hs
```diff
@@ -413,6 +442,32 @@ fromL2SatelliteObservation sat l1 l1e l2 l2e =
                                       -- pseudorange valid
     }
 
+-- | Validate IODE/IODC flags. The IODC and IODE flags (least significant bits) should be
+-- equal. IODC is Word16 while IODE is Word8, so we shift IODC to drop the most significant bits.
+-- We return a 1 if valid, 0 if invalid.
+validateIodcIode :: Word16 -> Word8 -> Word8
+validateIodcIode iodc iode =
+  let iodc' = fromIntegral $ iodc `shiftL` 8 `shiftR` 8 in
+  if  iodc' == iode then 1 else 0
```
mfine: nit - this file's used `where` clauses in the pure code.

haskell/src/Data/RTCM3/SBP.hs
```diff
@@ -544,6 +657,10 @@ convert = \case
     wn <- view storeWn
     liftIO $ writeIORef wn $ toWn $ m ^. msg1013_header ^. messageHeader_mjd
     return mempty
+  (RTCM3Msg1019 m _rtcm3) -> do
+    let sender = 0
+    m' <- fromMsg1019 m
+    return [SBPMsgEphemerisGps m' $ toSBP m' $ toSender sender]
```
mfine: if this is using sender = 0, then no need for the var or calling toSender:

```haskell
  return [SBPMsgEphemerisGps m' $ toSBP m' 0]
```

though it could be helpful to be consistent.

mfine:
tests look good to me! thanks!

PR #35, opened 2017-03-23
Round the phase value instead of truncating it

mfine:
Round the phase value instead of truncating it for more accuracy and resolution.

/cc @pgrgich @ljbade


mfine:
Fixes https://github.com/swift-nav/skylark-tales/issues/517.

mookerji:
Wow

PR #40, opened 2017-07-20
Support sender id from RTCM message station ID

anth-cole:
Currently, if we switch stations with the RTCM stream (for example with VRS) this wouldn't be captured in the sbp.


mfine:
To avoid conflicts with some low number sender id's that have or had special meaning (0, 42 I think) - we or'd in 0xf on the high nibble (e.g., |= 0xf000).

https://github.com/swift-nav/gnss-converters/blob/master/haskell/src/Data/RTCM3/SBP.hs#L479

anth-cole:
@mfine thanks mark, didn't realize we had special sender ID's, added the change.

PR #41, opened 2017-07-28
Handle glo day rollover

anth-cole:
When we're within 3 hours of the end of the day, we're getting a negative value, this wraps the tod to be positive.


ljbade:
Sorry to get to this a bit late, but there are a few missing spaces with some of the if statements

PR #45, opened 2017-08-15
Minor reshuffling of state fields and initialization

mfine:
Pull out a field and go directly to the buffer. Also, initialize buffer and sender id. Add some more safety initializations just in case.

/cc @anth-cole


mfine:
Going to bring this in, tag it, and bring in to buildroot.

PR #47, opened 2017-08-24
Protect time differencing

mfine:
This should help protect the dow rollover. Not sure it's necessary with the double's being used here, but should help for clarity. See [rfc1982](https://tools.ietf.org/html/rfc1982#section-3.2), though not sure it's relevant here.


anth-cole:
lgtm - helps clarity I think

PR #53, opened 2017-09-21
Reject any RTCM obs that have the same timestamp as one already received

anth-cole:
This keeps track of the last timestamp of the RTCM obs received and doesn't accept following RTCM obs messages of same or different type that have the same timestamp.

This is a problem if we get a stream with 1002 and 1004 messages, that are the L1 only and L1/L2 RTCM messages. While this is not a valid input stream, this will crash the receiver currently, this handles the situation more gracefully by ignoring any subsequent messages with identical RTCM time of week.

### DIFF
--- a/c/include/rtcm3_sbp.h
+++ b/c/include/rtcm3_sbp.h
@@ -18,7 +18,7 @@
 
 #define MAX_OBS_PER_EPOCH 56
 /* MAX valid value (ms) for GPS is 604799999 and GLO is 86401999 */
-#define INVALID_TOW 0xFFFF
+#define INVALID_TIME 0xFFFF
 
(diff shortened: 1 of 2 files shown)


anth-cole:
Bench tested on a septentrio mountpoint sending 1002 and 1004 messages and piksi operates fine.

benjamin0:
If possible, I'd add a print statement so that users know they have a bad RTCM config. Otherwise, LGTM

anth-cole:
Discussed in triage, verdict was that no information logging was necessary considering the amount of bad RTCM config options already exist (no warning for unsupported messages etc) going to address this in a later FW release.

PR #54, opened 2017-09-22
Add sanity check for base observations

silverjam:
If the base observations start getting too old to be used (> 10s) --
invoke an (optionally provided) callback to indicate this.

### DIFF
--- a/c/src/rtcm3_sbp.c
+++ b/c/src/rtcm3_sbp.c
@@ -19,7 +19,7 @@
 static void validate_base_obs_sanity(struct rtcm3_sbp_state *state, gps_time_sec_t *obs_time, const gps_time_sec_t *rover_time);
 
 void rtcm2sbp_init(struct rtcm3_sbp_state *state,
-                   void (*cb)(u8 msg_id, u8 length, u8 *buffer, u16 sender_id),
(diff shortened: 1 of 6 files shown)


c/src/rtcm3_sbp.c
```diff
@@ -28,6 +31,7 @@ void rtcm2sbp_init(struct rtcm3_sbp_state *state,
 
   state->sender_id = 0;
   state->cb = cb;
```
anth-cole: if we're going to add a second call back, should we make this one a more descriptive name? Something like `cb_rtcm_to_sbp` or something?

c/src/rtcm3_sbp_internal.h
```diff
@@ -16,6 +16,8 @@
 #include <rtcm3_messages.h>
 #include <rtcm3_sbp.h>
 
+#include <libsbp/piksi.h>
```
anth-cole: do we need this for something?

c/src/rtcm3_sbp.c
silverjam: fixed

c/src/rtcm3_sbp_internal.h
silverjam: removed

PR #55, opened 2017-11-21
Add glo bias msg support

anth-cole:
Add support for the RTCM 1230 message to support GLO IFB and glonass ambiguity fixing.

@mfine we'll need to add this to the haskell convertor as well and support it on the skylark endpoints in order to support glonass fixing.


mfine:
Sounds good. Will work on it.

PR #57, opened 2017-11-28
Need to account for the sign

anth-cole:
@scarcanague @mfine 

Add the interpretation of the bias sign

### DIFF
--- a/c/src/rtcm3_sbp.c
+++ b/c/src/rtcm3_sbp.c
@@ -15,7 +15,6 @@
 #include <math.h>
 #include <string.h>
 #include <assert.h>
-#include <rtcm3_messages.h>
 
 static void validate_base_obs_sanity(struct rtcm3_sbp_state *state, gps_time_sec_t *obs_time, const gps_time_sec_t *rover_time);
 
--- a/c/src/rtcm3_sbp.c
+++ b/c/src/rtcm3_sbp.c
@@ -15,6 +15,7 @@
 #include <math.h>
 #include <string.h>
 #include <assert.h>
+#include <rtcm3_messages.h>
 
 static void validate_base_obs_sanity(struct rtcm3_sbp_state *state, gps_time_sec_t *obs_time, const gps_time_sec_t *rover_time);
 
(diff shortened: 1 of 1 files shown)


mfine:
:+1:

c/src/rtcm3_sbp.c
```diff
@@ -451,16 +451,18 @@ void rtcm3_1230_to_sbp(const rtcm_msg_1230 *rtcm_1230,
 {
   sbp_glo_bias->mask = rtcm_1230->fdma_signal_mask;
   /* Resolution 2cm */
-  sbp_glo_bias->l1ca_bias = rtcm_1230->L1_CA_cpb_meter * 50;
-  sbp_glo_bias->l1p_bias = rtcm_1230->L1_P_cpb_meter * 50;
-  sbp_glo_bias->l2ca_bias = rtcm_1230->L2_CA_cpb_meter * 50;
-  sbp_glo_bias->l2p_bias = rtcm_1230->L2_P_cpb_meter * 50;
+  s8 sign_indicator = rtcm_1230->bias_indicator == 0 ? 1 : -1;
```
nsirola: This looks strange to be. 
If the indicator is 0 then the phase range can be aligned with these corrections, yes.
But if they are already aligned (indicator=1), why would we want to unalign them by applying reverse corrections?

anth-cole: The standard's a bit unclear to me, but here's the relevant fields

image

Looking at the data field nots for DF423 it seems DF421 should apply the bias in the opposite direction if 0 correct?

nsirola: Yes, but only if we wanted to un-align the already aligned observations. In SBP case, we should be happy with the already aligned observations, and either not send the GLO phases message at all, or send it with zero values, so receiver does not apply any more offsets to the already aligned offsets?

anth-cole: oh wow - you're right, I've been reading the standard completely wrong. Absolutely, we should set the phase values to zero, change the indicator, and send the SBP message

nsirola: Pull request at https://github.com/swift-nav/gnss-converters/pull/135

PR #59, opened 2017-12-23
Add 1033 support

anth-cole:
This adds support for the 1033 message, to allow us to construct a glonass bias message from the receiver type field. I've also added a timeout so this message will only be used if we haven't seena  1230 message in the last 45 seconds.

I still need to ensure I've used the correct IGS receiver identifiers (the strong identifying the receiver manufacturer)  as well as check the bias values being encoded, but I think the data flow should work

@silverjam - would you mind taking a look over this?

@scarcanague do these values align with what you've seen? Some of them (The nov and leica ones) are definitely different to what I remember they should be. I'm inclined to try and find some streams with both 1033 and 1230 supported and align that way to what's in the wild.

All rcv callsigns come from here ftp://igs.org/pub/station/general/rcvr_ant.tab


c/src/rtcm3_sbp.c
```diff
@@ -446,27 +459,61 @@ void sbp_to_rtcm3_1006(const msg_base_pos_ecef_t *sbp_base_pos,
...
+void rtcm3_1033_to_sbp(const rtcm_msg_1033 *rtcm_1033,
+                       msg_glo_biases_t *sbp_glo_bias)
+{
+  sbp_glo_bias->mask = 0;
+  /* Resolution 2cm */
+  if(strstr(rtcm_1033->rcv_descriptor,"TRIMBLE" ) != NULL || strstr(rtcm_1033->rcv_descriptor,"ASHTECH") != NULL) {
+    sbp_glo_bias->mask = 9;
+    sbp_glo_bias->l1ca_bias = 18.8 * GLO_BIAS_RESOLUTION;
+    sbp_glo_bias->l2p_bias = 23.2 * GLO_BIAS_RESOLUTION;
+  } else if( strstr(rtcm_1033->rcv_descriptor,"LEICA") != NULL || strstr(rtcm_1033->rcv_descriptor,"NOV") != NULL
```
scarcanague: Are you sure Leica is not called LEI sometimes? Same for Ashtech "ASH"? It is also missing SOKKIA I think which are the same than  novatel

c/src/rtcm3_sbp.c
```diff
@@ -446,27 +459,61 @@ void sbp_to_rtcm3_1006(const msg_base_pos_ecef_t *sbp_base_pos,
...
+    sbp_glo_bias->l2p_bias = 0.0 * GLO_BIAS_RESOLUTION;
+  } else if( strstr(rtcm_1033->rcv_descriptor,"TOP") != NULL) {
+    sbp_glo_bias->mask = 9;
+    sbp_glo_bias->l1ca_bias = -2.56 * GLO_BIAS_RESOLUTION;
+    sbp_glo_bias->l2p_bias = 3.74 * GLO_BIAS_RESOLUTION;
+  } else if( strstr(rtcm_1033->rcv_descriptor,"HEM") != NULL) {
+    sbp_glo_bias->mask = 9;
+    sbp_glo_bias->l1ca_bias = -0.3 * GLO_BIAS_RESOLUTION;
+    sbp_glo_bias->l2p_bias = 3.5 * GLO_BIAS_RESOLUTION;
+  } /*else if( strstr(rtcm_1033->rcv_descriptor,"JAVAD") != NULL) {
```
scarcanague: I think it should also contain the JPS tag

c/src/rtcm3_sbp.c
```diff
@@ -446,27 +459,61 @@ void sbp_to_rtcm3_1006(const msg_base_pos_ecef_t *sbp_base_pos,
...
+    Can add javad once we know the bias values
+    sbp_glo_bias->mask = 9;
+    sbp_glo_bias->l1ca_bias = -0.3 * GLO_BIAS_RESOLUTION;
+    sbp_glo_bias->l2p_bias = 3.5 * GLO_BIAS_RESOLUTION;
+  } */
+}
```
scarcanague: According to Skylark table, we may be also missing Novariant and Navcom (probably important for the ag business)

c/src/rtcm3_sbp.c
anth-cole: the antenna's are called LEI but the receviers names by IGS should be the full LEICA

c/src/rtcm3_sbp.c
anth-cole: JPS is typically a different bias level (or it can be I think). I'll uncomment this when I find a receiver to hook into and confirm the values

c/src/rtcm3_sbp.c
anth-cole: good point, will try and get those too

anth-cole:
@scarcanague I've only managed to find one trimble stream so far giving both the 1033 and the 1230 message, and the bias values it contains differs from what we see from geo++, which aligns with what I've seen previously. It's fine not to have the values, as you're right we'll go back to GPS only, but having the wrong values will be bad

scarcanague:
@anth-cole Hum strange, how different are the values? I f I remember well, every meter of difference translates in a maximum of 0.85mm difference at ambiguity level so there is definitely a tolerance

c/include/rtcm3_sbp.h
```diff
@@ -20,6 +20,12 @@
 /* MAX valid value (ms) for GPS is 604799999 and GLO is 86401999 */
 #define INVALID_TIME 0xFFFF
 
+/* Multiplier for glonass bias resolution scaling */
+#define GLO_BIAS_RESOLUTION 50
+
+/* How long to wait after receiving a 1230 message before accepting the 1033 message again */
+#define MSG_1230_TIMEOUT 45
```
silverjam: What are the units of this constant?  Should add this into the comment or the naming of the constant.

c/src/rtcm3_sbp.c
silverjam: Can we put these magic numbers into constants and explain where they come from?  Something like:

```c
#define TRIMBLE_L1CA_BIAS 18.8
```

c/src/rtcm3_sbp.c
silverjam: Why is this code commented out?

c/include/rtcm3_sbp.h
anth-cole: good point - done, renamed to `MSG_1230_TIMEOUT_SEC`

anth-cole:
@silverjam thanks for the review, I've run clang-format over the repo, it does confuse this PR a bit with a lot of other formatting changes. Still to do is find data to support JAVAD, HEMISPHERE and TOPCON receivers.

silverjam:
Sorry, I should have only requested clang formatting on the changed code, but this looks good.

Why do we need to have different biases for each receiver?  I see that we don't have a logging/warning mechanism for this code (like a callback), maybe we should add one so we can log if we encounter a receiver we don't recognize?  That is, if it's a serious enough condition to justify a warning.

anth-cole:
@silverjam I thought about only clang'ing my change, but that actually made it worse overall I thought so may as well do the lot.

These biases represent the glonass bias between the code and phase observation that are unique to each receiver manufacturer and that we need to compensate in order to do GLO amb fixing. If we get a receiver string we don't recognise, we just don't send on the SBP_GLO_BIAS message - the console is planned to warn if this message is not being received, the piksi will run normally but GLO sats will not be passed to ambiguity resolution.

PR #62, opened 2018-01-10
Log RTCM Message Type 1029 Manually as SBP_MSG_LOG

benjaminaltieri:
Addressing swift-nav/piksi_v3_bug_tracking#921, and downstream on the changed supplied by PR https://github.com/swift-nav/librtcm/pull/30, this PR covers handling the RTCM 1029 message in `rtcm2sbp_decode_frame`. The current strategy is to place the content of the 1029 (unicode text string) directly into an sbp log message and use the sbp callback to emit the message onto the bus. 

Details:

 * Protoype for sbp callback was updated to reflect being a u16 (SBP_MSG_LOG > 255)
 * Temporarily checked librtcm out to https://github.com/swift-nav/librtcm/pull/30 this will be updated after that PR is merged to master (how do we normally handle dependecies like this?)
 * The max framing size for an sbp message and the max size of the unicode buffer supplied by the 1029 are the same, but the inclusion of the sbp log level effective truncates the 1029 range to 254. I've added a check for this but it's rather minimal and would not cover inspection of the unicode string to remove a full unicode character. May need to add as issue.

### DIFF
--- a/c/src/rtcm3_sbp.c
+++ b/c/src/rtcm3_sbp.c
@@ -13,6 +13,7 @@
 #include "rtcm3_sbp_internal.h"
 #include <assert.h>
 #include <libsbp/observation.h>
+#include <libsbp/logging.h>
 #include <math.h>
 #include <rtcm3_decode.h>
 #include <string.h>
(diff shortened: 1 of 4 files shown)


c/src/rtcm3_sbp.c
```diff
@@ -621,6 +619,8 @@ bool no_1230_received(struct rtcm3_sbp_state *state) {
   return false;
 }
 
+#define SBP_FRAMING_MAX_PAYLOAD_SIZE (255u)
```
silverjam: Can we put these up at the top of file please

benjaminaltieri:
Also good to merge.

PR #63, opened 2018-01-12
Anthony/additional rtcm functionality

anth-cole:
This PR got a bit bigger than first planned but does the following

- Finishes the 1033 message support
- Adds GEO++ streams to 1033 message support
- Adds a warning if we see an MSM message from GPS or GLO (once only)
- Refactors the unit test setup


c/src/rtcm3_sbp.c
```diff
@@ -622,52 +728,61 @@ bool no_1230_received(struct rtcm3_sbp_state *state) {
   return false;
 }
 
-void send_1029(rtcm_msg_1029 *msg_1029, struct rtcm3_sbp_state *state){
+void send_1029(rtcm_msg_1029 *msg_1029, struct rtcm3_sbp_state *state) {
   uint8_t message[SBP_FRAMING_MAX_PAYLOAD_SIZE] = "RTCM: ";
```
benjaminaltieri: If this used a macro, we could link the string here to preamble size explicitly via:
```
#define RTCM_1029_LOG_MSG_PREAMBLE "RTCM: "
uint8_t message[SBP_FRAMING_MAX_PAYLOAD_SIZE] = RTCM_1029_LOG_MSG_PREAMBLE;
uint8_t preamble_size = sizeof(RTCM_1029_LOG_MSG_PREAMBLE) - 1; // minus one for the terminating zero
```

anth-cole: Awesome, thanks for that, I was looking for a way of linking the two and couldn't think of one, this is a lot better imo.

PR #66, opened 2018-03-06
Handle 1012 coming first

anth-cole:
There was a problem where the de-duplication logic, to ensure we only process 1 GPS message or 1 GLO message in an epoch where the GLO test was using GPS state. This was fine if the GPS message came first, but caused a problem where we would dump the 1004 message if the 1012 message came first.


mfine:
This de-duping looks good :+1: - easier to follow than before.

PR #69, opened 2018-04-20
Add Check Test Framework

benjaminaltieri:
For C sources we would like to have a testing framework to organize testing as new converters are added (i.e. NMEA)

This handles modifying current CMakelists configuration and tests as written to pull in libcheck.

I've also borrowed a cmake script for pulling libsbp in as a library to take advantage of linking and other cmake niceties.


c/tests/check_rtcm3.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 Swift Navigation Inc.
+ * Copyright (C) 2018 Swift Navigation Inc.
```
anth-cole: We don't typically update the year when modifying the file, only on creation

c/tests/check_rtcm3.c
```diff
@@ -58,22 +65,23 @@ void sbp_callback_glo_day_rollover(u16 msg_id, u8 length, u8 *buffer,
...
     u8 num_sbp_msgs = msg->header.n_obs >> 4;
-    assert(num_sbp_msgs > 2);
+    ck_assert(num_sbp_msgs > 2);
   }
 }
 
 void check_biases(msg_glo_biases_t *sbp_glo_msg) {
   if (sbp_glo_msg->mask & 0x01) {
-    assert(sbp_glo_msg->l1ca_bias / GLO_BIAS_RESOLUTION == expected_L1CA_bias);
+    //ck_assert_double_eq(sbp_glo_msg->l1ca_bias / GLO_BIAS_RESOLUTION, expected_L1CA_bias);
```
anth-cole: nit: assume this was test code?

PR #70, opened 2018-04-24
[WIP] Introducing nmea module from PFWP with dependencies

benjaminaltieri:
In this case: 
- libswiftnav-common: provides commond gnss structures and utilities
 - libswiftnav/observation.h: modified to provide constant defines
 - PFWP/sbp_utils.h: some utilities used for sbp manipulation

These dependencies are subject to change depending on the depth of
refactor to be performed in the nmea generation APIs. The goal of
this commit was to show that the code can be compiled and exercised
outside of PFWP.

The one test (GGA) for now is not functional. Tests will be added
individually for each case to be validated in code review.

DEVC-457 #in-review


c/refactor/common/tests/check_main.c
```diff
@@ -0,0 +1,38 @@
...
+  srunner_add_suite(sr, coord_system_suite());
+  srunner_add_suite(sr, linear_algebra_suite());
+  srunner_add_suite(sr, troposphere_suite());
+  srunner_add_suite(sr, ephemeris_suite());
+  srunner_add_suite(sr, set_suite());
+  srunner_add_suite(sr, gnss_time_test_suite());
+  srunner_add_suite(sr, signal_test_suite());
+  srunner_add_suite(sr, glo_map_test_suite());
+  srunner_add_suite(sr, shm_suite());
+  //srunner_add_suite(sr, pvt_test_suite());
```
benjaminaltieri: This test segfaults when built and run with travis. Unable to reproduce locally and haven't been able to root cause since pushing changes to travis is the only way I can test this currently.

denniszollo:
Why do we need so much from libswiftnav?  

Are those dependencies required to calculate azimuth and elevation?  If so, I don't think the converter code should be doing that and that should be handled with an az and el SBP message that is missing from the protocol at present.

benjaminaltieri:
As a starting point I decided to copy the common section of the repository whole-hog since it already has build and test support via cmake. Doing this is definitely overkill and to the point of your question so far the only functional requirements I have seen when I exercise the individual NMEA generation routines are those that involve GPS -> UTC time conversion. Until I get deeper into the refactor I can't say for sure though, and I need to work out a detailed map of the functional dependencies from libswiftnav-common to facilitate the discussion.

anth-cole:
closing as done in other PR

PR #73, opened 2018-05-17
EX-106: multi-constellation MSM support

nsirola:
Implement support for RTCM MSM messages 4-7 for all constellations, and moves `librtcm` pointer to https://github.com/swift-nav/librtcm/pull/37 to include MSM decoding support.

If MSM message is succesfully parsed, ignores legacy observation messages for the next 60 seconds.

Note that this initial implementation passes through all the signals as they are read from RTCM (given that there is a code enum defined in LNSP for that signal), whether or not PM can make use of that signal. There is a separate issue for filtering and mapping the signals into ones usable by Piksi: https://swift-nav.atlassian.net/browse/EX-139

Sanitizes `memset` usage also for old code.

Also adds CRC check in the unit tests when reading RTCM messages from a file, there were quite lot of garbage being parsed and updating time stamps with bogus values.


c/src/rtcm3_sbp.c
```diff
@@ -353,16 +353,12 @@ void add_obs_to_buffer(const rtcm_obs_message *new_rtcm_obs,
...
-  const uint32_t header_size = sizeof(observation_header_t);
-  const uint32_t obs_size = sizeof(packed_obs_content_t);
-  const uint32_t max_obs_in_sbp =
-      ((SBP_FRAMING_MAX_PAYLOAD_SIZE - header_size) / obs_size);
-
   /* We want the ceiling of n_obs divided by max obs in a single message to get
    * total number of messages needed */
   const u8 total_messages =
-      1 + ((sbp_obs_buffer->header.n_obs - 1) / max_obs_in_sbp);
+      1 + ((sbp_obs_buffer->header.n_obs - 1) / MAX_OBS_IN_SBP);
```
adel-mamin: Maybe it would also make sense to add 
`assert(sbp_obs_buffer->header.n_obs != 0);`

nsirola: I did and found out there are sometimes empty message that are cleaner to just let pass through as they skip the loop below in any case.

c/src/rtcm3_sbp.c
```diff
@@ -207,6 +258,11 @@ void rtcm2sbp_decode_frame(const uint8_t *frame,
 
 void add_glo_obs_to_buffer(const rtcm_obs_message *new_rtcm_obs,
                            struct rtcm3_sbp_state *state) {
+  if (state->using_msm) {
```
anth-cole: I'd do this in a later PR, but I'd suggest instead of storing this as a bool, we store the timestamp of the last MSM epoch and then we switch back to legacy after 60s.

nsirola: Done

c/src/rtcm3_sbp.c
```diff
@@ -526,8 +526,13 @@ void rtcm3_to_sbp(const rtcm_obs_message *rtcm_obs,
 
         sbp_freq->sid.sat = rtcm_obs->sats[sat].svId;
         if (gps_obs_message(rtcm_obs->header.msg_num)) {
-          sbp_freq->sid.code =
-              get_gps_sbp_code(freq, rtcm_obs->sats[sat].obs[freq].code);
+          if (sbp_freq->sid.sat < 40) {
```
ljbade: I don't think this is correct the RTCM spec says >32 is for SBAS, but starts counting at 40. I think we should just ignore PRNs in the range 33-39.

ljbade:
@nsirola I think there is similar PRN issue with GLONASS in 1012 messages, it is also define 1-24 for GLO and >32 for SBAS. I think we should just ignore any PRN > 24.

ljbade:
We should also ignore PRN 0 for both systems.

nsirola:
Added an issue https://swift-nav.atlassian.net/browse/EX-146 for checking PRN ranges in MSM parsing.

ljbade:
@nsirola the latest change LGTM

ljbade:
LGTM again

I haven't observed any issues yet from running this for a few hours

nsirola:
Ok, will leave this running overnight and merge in the morning if there are no further issues.

PR #77, opened 2018-05-31
MSM fixes

mfine:
Fixes to the haskell conversions:

+ cell mask had an off by one error.
+ flipped the hca bit
+ had the pseudorange unit in the divisor
+ 1/4 cnr values

/cc @nsirola @benjamin0 @ljbade


ljbade:
When I get home next week, ill test out this new code against the local stations to verify I get RTK fixes.

PR #79, opened 2018-06-01
Add more range checks to satellite IDs and GLONASS FCNs

ljbade:
Add more checks around valid satellite IDs and GLONASS FCN slots.

This matches what we check in the C conversion.

### DIFF
--- a/haskell/src/Data/RTCM3/SBP/Observations.hs
+++ b/haskell/src/Data/RTCM3/SBP/Observations.hs
@@ -182,7 +182,7 @@ toGpsL1PackedObsContents sat l1o l1eo
       , _packedObsContent_flags = obsFlags
       }
   where
-    no = sat > gpsMaxSat || l1o ^. gpsL1Observation_pseudorange == 524288 || l1o ^. gpsL1Observation_carrierMinusCode == -524288
+    no = sat < 1 || sat > gpsMaxSat || l1o ^. gpsL1Observation_pseudorange == 524288 || l1o ^. gpsL1Observation_carrierMinusCode == -524288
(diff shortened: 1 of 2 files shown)


haskell/src/Data/RTCM3/SBP/Observations.hs
```diff
@@ -222,7 +222,7 @@ toGlonassL1PackedObsContents sat l1o l1eo
         , _packedObsContent_flags = obsFlags
         }
   where
-    no = sat > glonassMaxSat || l1o ^. glonassL1Observation_carrierMinusCode == -524288
+    no = sat < 1 || sat > glonassMaxSat || l1o ^. glonassL1Observation_frequency < 0 || l1o ^. glonassL1Observation_frequency > 13 || l1o ^. glonassL1Observation_carrierMinusCode == -524288
```
mfine: just a small nit here - `glonassL1Observation_frequency` is a Word value and can't be under 0.

mfine: Also, it's a shifted value (+7) - so check should be > 20, not > 13.

ljbade: Ah I couldn't remember if Word was signed or not.

I think the 7 is subtracted later, see `glonassL1CarrierPhase`. This matches the DF419 encoding in RTCM table 3.4-6

mfine: The 7 is subtracted later in `glonassL1CarrierPhase`:

https://github.com/swift-nav/gnss-converters/blob/master/haskell/src/Data/RTCM3/SBP/Observations.hs#L84

The check here against `glonassL1Observation_frequency` is an unshifted value - it has values from 0 - 20. So if you check for values greater than 13, you will exclude frequencies 14-20 which get later mapped to 7-13.

See the fix here: https://github.com/swift-nav/gnss-converters/pull/81

PR #81, opened 2018-06-02
MSM validity checks

mfine:
Add checks on data.

/cc @ljbade

### DIFF
--- a/haskell/src/Data/RTCM3/SBP/Observations.hs
+++ b/haskell/src/Data/RTCM3/SBP/Observations.hs
@@ -222,7 +222,7 @@ toGlonassL1PackedObsContents sat l1o l1eo
         , _packedObsContent_flags = obsFlags
         }
   where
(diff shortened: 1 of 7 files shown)


haskell/src/Data/RTCM3/SBP/Observations.hs
```diff
@@ -222,7 +222,7 @@ toGlonassL1PackedObsContents sat l1o l1eo
         , _packedObsContent_flags = obsFlags
         }
   where
-    no = sat < 1 || sat > glonassMaxSat || l1o ^. glonassL1Observation_frequency < 0 || l1o ^. glonassL1Observation_frequency > 13 || l1o ^. glonassL1Observation_carrierMinusCode == -524288
+    no = sat < 1 || sat > glonassMaxSat || l1o ^. glonassL1Observation_frequency > 20 || l1o ^. glonassL1Observation_carrierMinusCode == -524288
```
ljbade: I still think this should be >13, see RTCM 3.3 spec Table 3.4-6 page 107/108

ljbade: nvm I realise now I'm getting mixed up between DF040 from 1012 message and DF419 from MSM

PR #83, opened 2018-06-04
EX-155: Add GLO FCN map and pass it to MSM4/6 decoders

nsirola:
Add the GLO FCN map into state, a callback function for the `sbp_rtcm3_bridge` in buildroot to update it with, and move `librtcm` pointer to https://github.com/swift-nav/librtcm/pull/42.

# Testing

Sanity checks with stored MSM logs mainly to check that this does not crash. 

Buildroot PR using this fixed into a live MSM4 stream https://github.com/swift-nav/piksi_buildroot/pull/643

### DIFF
--- a/c/src/rtcm3_sbp.c
+++ b/c/src/rtcm3_sbp.c
@@ -15,6 +15,7 @@
 #include <math.h>
 #include <rtcm3_decode.h>
 #include <rtcm3_msm_utils.h>
+#include <stdio.h>
 #include <string.h>
 #include "rtcm3_sbp_internal.h"
 
@@ -47,6 +48,10 @@ void rtcm2sbp_init(
   state->last_msm_received.tow = 0;
 
   state->sent_msm_warning = false;
+
+  for (u8 i = 0; i < NUM_GLO_MAP_INDICES; i++) {
+    state->glo_sv_id_fcn_map[i] = MSM_GLO_FCN_UNKNOWN;
+  }
 
   memset(state->obs_buffer, 0, OBS_BUFFER_SIZE);
 }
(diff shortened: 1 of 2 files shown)


c/src/rtcm3_sbp.c
```diff
@@ -783,6 +790,18 @@ void rtcm2sbp_set_leap_second(s8 leap_seconds, struct rtcm3_sbp_state *state) {
...
 }
 
+void rtcm2sbp_set_glo_fcn(sbp_gnss_signal_t sid,
+                          u8 sbp_fcn,
+                          struct rtcm3_sbp_state *state) {
+  /* convert FCN from SBP representation to RTCM representation */
+  if (SBP_GLO_FCN_UNKNOWN == sbp_fcn) {
+    state->glo_sv_id_fcn_map[sid.sat] = MSM_GLO_FCN_UNKNOWN;
+  } else {
+    s16 fcn = sbp_fcn - SBP_GLO_FCN_OFFSET;
```
pmiettinen: What happens here, why `s16`?

nsirola: Being paranoid about truncating negative values. In suppose `s8` would have been enough.

pmiettinen: ok, well the worst that can happen on the negative side is `0 - SBP_GLO_FCN_OFFSET = -SBP_GLO_FCN_OFFSET`

Bigger problem would be if `sbp_fcn > SCHAR_MAX + SBP_GLO_FCN_OFFSET`.

c/src/rtcm3_sbp.c
```diff
@@ -783,6 +790,18 @@ void rtcm2sbp_set_leap_second(s8 leap_seconds, struct rtcm3_sbp_state *state) {
...
+    state->glo_sv_id_fcn_map[sid.sat] = (u8)fcn + MSM_GLO_FCN_OFFSET;
```
pmiettinen: This is incorrect now because cast operator precedes addition. Consider `fcn = -1` which will be casted into `255`.

nsirola: Fixed. (And makes one wonder how it worked at all)

pmiettinen: One more thing. If my research is correct, `MSM_GLO_FCN_OFFSET = 7` and `SBP_GLO_FCN_OFFSET = 8` .

If `sbp_fcn` can be `0`:
1. fcn = 0 - 8
2. -8 + 7 = -1
And map is assigned with 255.

nsirola: `sbp_fcn` value 0 is the unknown value `SBP_GLO_FCN_UNKNOWN` and it's checked separately first

PR #86, opened 2018-06-13
EX-178 Handle GPS L1W and L2W in MSM

ljbade:


haskell/src/Data/RTCM3/SBP/MSM.hs
```diff
@@ -61,6 +61,7 @@ toGpsCode :: Word8 -> Maybe Word8
 toGpsCode sig
   | sig == 2  = Just 0
   | sig == 3  = Just 5
+  | sig == 4  = Just 5
```
mfine: So this is annoying how it works today (and should get cleaned up), but you also have to add the signal to the frequency call `toGpsFrequency`.

benjamin0:
Sounds fine to me, but not sure if any functionality will need to change downstream such as the code mapper. cc @scarcanague?

scarcanague:
The risk of that approach is if the stream contains both L2W and L2P signals, they will be concatenated and potentially mixed with no warning. This may not be a use case in practice for the stations

ljbade:
@scarcanague This would be a possibility already with the 1004 messages.

ljbade:
@mfine added toGpsFrequency, please review again

ljbade:
Should mention Stefan tested this code for me in Japan and he got RTK fix with it.

PR #87, opened 2018-06-19
[WIP] Warn on L1P/L2P when converting RTCM-1012 and MSM

benjaminaltieri:
~It looks like the tests rely on these L1P and L2P signals not being rejected, so at this point the tests are failing. I put this PR up to get feedback and see if maybe this isn't the right approach to solve our 'unsupported' codes problem~

Updated to retain behavior and only emit logs.

RE: [DEVC-767](https://swift-nav.atlassian.net/browse/DEVC-767)

### DIFF
--- a/c/src/rtcm3_sbp.c
+++ b/c/src/rtcm3_sbp.c
@@ -48,6 +48,9 @@ void rtcm2sbp_init(
   state->last_msm_received.tow = 0;
 
   state->sent_msm_warning = false;
+  for(u8 i = 0; i < UNSUPPORTED_CODE_MAX; i++) {
+    state->sent_code_warning[i] = false;
+  }
 
   for (u8 i = 0; i < GLO_LAST_PRN + 1; i++) {
     state->glo_sv_id_fcn_map[i] = MSM_GLO_FCN_UNKNOWN;
@@ -499,15 +502,29 @@ code_t get_gps_sbp_code(u8 freq, u8 rtcm_code) {
   return code;
 }
 
-code_t get_glo_sbp_code(u8 freq, u8 rtcm_code) {
-  (void)rtcm_code;
+code_t get_glo_sbp_code(u8 freq,
+                        u8 rtcm_code,
+                        struct rtcm3_sbp_state *state) {
   code_t code = CODE_INVALID;
   if (freq == L1_FREQ) {
-    code = CODE_GLO_L1OF;
-    /* CODE_GLO_L1P currently not supported in sbp */
+    if (rtcm_code == 0) {
+      code = CODE_GLO_L1OF;
+    } else {
+      /* CODE_GLO_L1P currently not supported in sbp */
+      send_unsupported_code_warning(UNSUPPORTED_CODE_GLO_L1P, state);
+      code = CODE_INVALID;
+    }
   } else if (freq == L2_FREQ) {
-    code = CODE_GLO_L2OF;
-    /* CODE_GLO_L2P currently not supported in sbp */
+    if (rtcm_code == 0) {
+      code = CODE_GLO_L2OF;
+    } else if (rtcm_code == 1) {
+      /* CODE_GLO_L2P currently not supported in sbp */
+      send_unsupported_code_warning(UNSUPPORTED_CODE_GLO_L2P, state);
+      code = CODE_INVALID;
+    } else {
+      /* rtcm_code == 2 or 3 is reserved */
+      code = CODE_INVALID;
+    }
   }
   return code;
 }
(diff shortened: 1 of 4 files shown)


c/src/rtcm3_sbp.c
```diff
@@ -945,6 +968,25 @@ void send_buffer_full_error(const struct rtcm3_sbp_state *state) {
...
+
+void send_unsupported_code_warning(const unsupported_code_t unsupported_code, struct rtcm3_sbp_state *state) {
+  assert(unsupported_code < UNSUPPORTED_CODE_MAX);
+  if (!state->sent_code_warning[unsupported_code]) {
+    /* Only send 1 warning */
+    state->sent_code_warning[unsupported_code] = true;
+    uint8_t msg[CODE_WARNING_BUFFER_SIZE];
+    size_t count = snprintf((char *)msg, CODE_WARNING_BUFFER_SIZE, CODE_WARNING_FMT_STRING, unsupported_code_desc[unsupported_code]);
+    assert(count < CODE_WARNING_BUFFER_SIZE);
+    send_sbp_log_message( RTCM_CODE_LOGGING_LEVEL, msg, count, 0, state);
```
ljbade: extra space after (

c/src/rtcm3_sbp.c
```diff
@@ -48,6 +48,9 @@ void rtcm2sbp_init(
   state->last_msm_received.tow = 0;
 
   state->sent_msm_warning = false;
+  for(u8 i = 0; i < UNSUPPORTED_CODE_MAX; i++) {
```
ljbade: missing space before (

c/src/rtcm3_sbp.c
```diff
@@ -1041,21 +1083,25 @@ void add_msm_obs_to_buffer(const rtcm_msm_message *new_rtcm_obs,
...
+                             sbp_gnss_signal_t *sid,
+                             struct rtcm3_sbp_state *state) {
   code_t code = msm_signal_to_code(header, signal_index);
   u8 sat = msm_sat_to_prn(header, satellite_index);
   if (CODE_INVALID != code && PRN_INVALID != sat) {
     sid->code = code;
     sid->sat = sat;
     return true;
   } else {
+    if(CODE_INVALID == code) {
```
ljbade: add space before (

c/src/rtcm3_sbp.c
```diff
@@ -1041,21 +1083,25 @@ void add_msm_obs_to_buffer(const rtcm_msm_message *new_rtcm_obs,
...
+      send_unsupported_code_warning(UNSUPPORTED_CODE_UNKNOWN, state);
```
ljbade: it would be nicer if we could just have specific warning for L1P/L2P but is difficult because the code conversion is done in librtcm for MSM...

ljbade:
@anth-cole it was only L2P, no L2OF. I guess we can do the same for MSM, but need to keep in mind that if we somehow get both L2P and L2OF (for example Trimble tracks both so theoretically it could happen in the future) Startling will probably assert.

I think there was also concern when I mentioned this L2P stuff that there might be code biases or something?

benjaminaltieri:
Thanks for the PR review @anth-cole @ljbade I've modified to keep the previous behavior.

Mainly I want to make sure that 1) a single warning is sufficient and 2) having a generic warning for MSM will suffice. Otherwise I would either have to add a logging callback framework to `librtcm` to warn on L*P signals, or add L1P/L2P defines to `libsbp`, both of which would expand the scope of this issue quite a bit.

anth-cole:
@benjaminaltieri thanks for jumping on this, I think those assumptions are absolutely fine. We shouldn't invest too much in this I think

ljbade:
@anth-cole @benjamin0 is there also going to be a PR for librtcm to ensure we encode MSM GLO L1P/L2P as L1OF/L2OF?

benjaminaltieri:
I have no idea what the side-effects would be so I'll defer to @anth-cole on this. Maybe a new bug should be created to cover this specific behavior change since it's kind of piggybacking on [DEVC-767](https://swift-nav.atlassian.net/browse/DEVC-767)

scarcanague:
@ljbade Im a little confused, why do we need to change librtcm? Doesn't this PR ensure that no input stream will be GLO L1P/L2P?

scarcanague:
I see, that does make sense to change this in librtcm to mimic the behavior. @ljbade Do you think you can do that?

benjaminaltieri:
@scarcanague I'm happy to do it but it just wasn't clear to me what the consensus was. I'll put up a PR for @ljbade to review and update this PR to pull in those changes

PR #88, opened 2018-06-22
Streamline SBP messages

anth-cole:
Suppress the encoding into SBP of any observation types that are currently not supported on piksi.

@scarcanague @ljbade

### DIFF
--- a/c/src/rtcm3_sbp.c
+++ b/c/src/rtcm3_sbp.c
@@ -1122,8 +1122,6 @@ bool unsupported_signal(sbp_gnss_signal_t* sid) {
     case CODE_GAL_E6B:
     case CODE_GAL_E6C:
     case CODE_GAL_E6X:
-    case CODE_GAL_E7Q:
-    case CODE_GAL_E7X:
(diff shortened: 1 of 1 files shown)


c/src/rtcm3_sbp.c
```diff
@@ -1114,6 +1114,33 @@ static bool get_sid_from_msm(const rtcm_msm_header *header,
...
 
+bool unsupported_signal(sbp_gnss_signal_t* sid) {
+  switch(sid->code) {
+    case CODE_GPS_L5I:
+    case CODE_GPS_L5X:
+    case CODE_GPS_L5Q:
+    case CODE_GAL_E6B:
+    case CODE_GAL_E6C:
+    case CODE_GAL_E6X:
+    case CODE_GAL_E7Q:
```
scarcanague: As discussed on Slack, I think we should keep `CODE_GAL_E7Q` and `CODE_GAL_E7X' as they might be the only code sent by the base. We have some code in the firmware to then convert them to` CODE_GAL_E7I’

PR #89, opened 2018-06-25
Mark GLO L1P and L2P as invalid codes

ljbade:


nsirola:
Compiled this into the buildroot and and this is at least much better. Still getting occasional duplicate observations though (as shown by https://github.com/swift-nav/libswiftnav-private/pull/1549), but they come also from GPS so that might be a separate issue.

c/src/rtcm3_sbp.c
```diff
@@ -513,16 +513,15 @@ code_t get_glo_sbp_code(u8 freq,
     if (rtcm_code == 0) {
       code = CODE_GLO_L1OF;
     } else {
-      /* CODE_GLO_L1P currently not supported in sbp */
-      /* warn and then replace with a supported code for now */
+      /* CODE_GLO_L1P currently not supported in sbp so warn */
       send_unsupported_code_warning(UNSUPPORTED_CODE_GLO_L1P, state);
-      code = CODE_GLO_L1OF;
+      code = CODE_INVALID;
```
anth-cole: Why would we not map L1P to L1OF?

ljbade:
@anth-cole good point, reverted the change to 1012. This will suffice as a temporary fix in 1.6 until we can handle this properly in 1.7

PR #90, opened 2018-06-27
EX-206: Always check msm multiple flag

nsirola:
If the last MSM message has invalid time, its `multiple` field is not looked at, and the observation set is not sent out which caused some peculiarities:
- if the last message is missed, the observations are only when receiving the next epoch's first message, causing 1s lag in base observations
- sometimes it seems some of the collected measurements were lost altogether

(the actual change is in the commit https://github.com/swift-nav/gnss-converters/pull/90/commits/9562a9b6b6d463bd9b376932922939b9bc548d75, the other is just clang formatting)


nsirola:
Also fixed a bug in skipping unsupported signals where cell indexing for that message went out of sync.

nsirola:
The cell indexing bug was introduced on Monday with https://github.com/swift-nav/gnss-converters/pull/88 and looks to be the root cause of at least https://swift-nav.atlassian.net/browse/EX-207 and https://swift-nav.atlassian.net/browse/MAP-361

nsirola:
One more commit to address the case where SBAS or QZSS message is the last in sequence and it was skipped altogether.

c/src/rtcm3_sbp.c
```diff
@@ -237,13 +237,16 @@ void rtcm2sbp_decode_frame(const uint8_t *frame,
...
     case 1106:
     case 1107:
-      /* SBAS messages suppressed for now */
-      break;
     case 1114:
     case 1115:
     case 1116:
     case 1117:
-      /* QZSS messages suppressed for now */
+      /* SBAS and QZSS messages suppressed for now. Check only the Multiple
```
ljbade: I think we actually need to handle this for all messages in range 1070-1229 as per RTCM 3.3 section 3.5.16.3.8 Multiple MSM Output:
"Multiple MSM messages may be transmitted for a given physical epoch. In order to indicate the “End of Epoch” DF393 (MSM
Multiple Message Bit) has been introduced. This DF is the same for all GNSS and should be set to 1 if more MSM for same or another
GNSS follows for given physical epoch and given Reference Station ID.

In order to effectively utilize this bit, the decoding software should be aware of all possible GNSS transmitted in the stream. Since
new MSM messages for more GNSS are expected to be designed in the future, the following message range is reserved exclusively for
MSM
MT1070 – MT1229
It is guaranteed that 55-th bit (counting from one) represents MSM Multiple Message Bit in messages MTxxx1, MTxxx2, MTxxx3,
MTxxx4, MTxxx5, MTxxx6 and MTxxx7 from the above mentioned range of messages.
This enables the decoding software to detect the end of epoch without knowledge of the content or format of any new MSM for new
GNSS."

ljbade:
Tested the build-root build and confirm it fixes EX-207

c/src/rtcm3_sbp.c
nsirola: Refactored the sending of MSM obs buffer to be done separately for any MSM message type.

ljbade: Just tested the last build and everything appears to be working

PR #91, opened 2018-07-10
EX-230: Time diff handling fixes

nsirola:
- Fix `compute_glo_time` which returned invalid time stamps for 18 seconds before week roll over (tow overflow)
- Fix `compute_gps_time` which returned invalid Beidou time stamps for 3 hours before week roll over (tow overflow)
- Protect all calls to `gps_diff_time` with time validity checks
- Add a basic unit test for `compute_glo_time`
- Change all handling of integer second/millisecond timestamps to be done with integer data types


nsirola:
Noticed that the GLO time conversion will give whacky results during the leap second event, as well as on each midnight of the week after the leap second event (because the DOW computation from GPS TOW is off by one second), so added sanity check to invalidate GLO timestamp in this instance.

Again, even less facilities to test this in simulation.

c/src/rtcm3_sbp.c
```diff
@@ -62,9 +62,24 @@ void rtcm2sbp_init(
...
+static bool gps_time_valid(const gps_time_sec_t *t) {
+  return (t->wn != INVALID_TIME) && (t->tow < SEC_IN_WEEK);
+}
+
 static double gps_diff_time(const gps_time_sec_t *end,
                             const gps_time_sec_t *beginning) {
-  int week_diff = end->wn - beginning->wn;
+  assert(gps_time_valid(beginning));
+  assert(gps_time_valid(end));
+  s16 week_diff = end->wn - beginning->wn;
```
pmiettinen: `week_diff` is incorrectly interpreted as negative value if for example 
`end->wn > INT16_MAX` and `beginning->wn = 0`

c/src/rtcm3_sbp.c
```diff
@@ -818,8 +840,7 @@ void sbp_to_rtcm3_1230(const msg_glo_biases_t *sbp_glo_bias,
 
 void rtcm2sbp_set_gps_time(gps_time_sec_t *current_time,
                            struct rtcm3_sbp_state *state) {
-  state->time_from_rover_obs.wn = current_time->wn;
-  state->time_from_rover_obs.tow = current_time->tow;
+  state->time_from_rover_obs = *current_time;
   state->gps_time_updated = true;
```
pmiettinen: Consider using compiler barrier to prevent optimization actions reshuffling `gps_time_updated` flag setting to first. If single thread, then doesn't really matter.

c/src/rtcm3_sbp.c
```diff
@@ -818,8 +840,7 @@ void sbp_to_rtcm3_1230(const msg_glo_biases_t *sbp_glo_bias,
 
 void rtcm2sbp_set_gps_time(gps_time_sec_t *current_time,
```
pmiettinen: `const`?

c/src/rtcm3_sbp.c
```diff
@@ -861,15 +883,24 @@ void compute_gps_time(double tow_ms,
...
+  if (!state->leap_second_known) {
+    obs_time->wn = INVALID_TIME;
+    return;
+  }
   /* Need to work out DOW from GPS time first */
-  int rover_dow = rover_time->tow / SEC_IN_DAY;
-  int rover_tod = rover_time->tow - rover_dow * SEC_IN_DAY;
+  /* NOTE: this will produce incorrect day of week during the leap second, but
+   * the failure will be caught by the sanity check at the end */
+  u8 rover_dow = rover_time->tow / SEC_IN_DAY;
```
pmiettinen: Is the `rover_time` guaranteed normalized?

c/src/rtcm3_sbp.c
nsirola: Added a limit for week number in `gps_time_valid`. (we're still good until January 2608 though)

c/src/rtcm3_sbp.c
nsirola: Fixed by refactoring this boolean field away. This is the only place writing to `time_from_rover_obs`.

c/src/rtcm3_sbp.c
nsirola: Fixed

c/src/rtcm3_sbp.c
nsirola: We don't even decode the frame if it isn't. Added an assert nevertheless.

nsirola:
Fix to be verified next weekend before merging.

PR #96, opened 2018-08-03
WIP Anthony/add rtcm eph support

anth-cole:
So far no testing has been done - but the first WIP at adding GPS eph support. I'll add the other systems next but let me know if not having GPS tested is blocking anything

@mfine @mbavaro


mfine:
sweet - looking good. This is with https://github.com/swift-nav/librtcm/compare/master...anthony/add_eph_support? Just rough eyeballing, it seems like a lot of the numbers are coming over without being adjusted? Like, compare with: https://github.com/swift-nav/gnss-converters/blob/master/haskell/src/Data/RTCM3/SBP/Ephemerides.hs#L174-L197 Have'not looked too closely. Thanks for tackling this and helping us move on this!

c/src/rtcm3_sbp_ephemeris.c
```diff
@@ -0,0 +1,57 @@
...
+  sbp_gps_eph->common.ura = msg_eph->ura;
+  sbp_gps_eph->common.fit_interval = msg_eph->fit_interval;
+  sbp_gps_eph->common.valid = msg_eph->valid;
+  sbp_gps_eph->common.health_bits = msg_eph->health_bits;
+
+  sbp_gps_eph->tgd = msg_eph->kepler.tgd_gps_s * power2_31;
+
+  sbp_gps_eph->c_rs = msg_eph->kepler.crs * power2_5;
+  sbp_gps_eph->c_rc = msg_eph->kepler.crc * power2_5;
+  sbp_gps_eph->c_uc = msg_eph->kepler.cuc * power2_29;
```
anth-cole: @mfine I think most of these are converted here - it's a bit ugly at the moment. Thanks for the haskell link, I didn't realise I had that reference but that'll make checking this a lot easier.

mfine: ahh sweet, I think I missed the `power2_*` at the end - sorry about that.

c/src/rtcm3_sbp.c
```diff
@@ -178,6 +178,28 @@ void rtcm2sbp_decode_frame(const uint8_t *frame,
...
+                              (u8 *)&sbp_gps_eph,
+                              rtcm_2_sbp_sender_id(0));
+      }
+    }
+    case 1020: {
+      rtcm_msg_eph msg_eph;
+      if (RC_OK == rtcm3_decode_eph(&frame[byte], &msg_eph)){
+        msg_ephemeris_glo_t sbp_glo_eph;
+        rtcm3_glo_eph_to_sbp(&msg_eph, &sbp_glo_eph);
+        state->cb_rtcm_to_sbp(SBP_MSG_EPHEMERIS_GPS,
```
mbavaro: `SBP_MSG_EPHEMERIS_GLO` here?

c/src/rtcm3_sbp_ephemeris.c
```diff
@@ -0,0 +1,57 @@
+
+
+#include "rtcm3_sbp_internal.h"
+#include <math.h>
+
+#define power2_55 pow(2,-55) /* 2^-31 */
```
mbavaro: please let's replace this define with explicit numeric values.. `pow()` is really inefficient.. why adding unnecessary CPU cycles for a constant?
`libswiftnav` has a bunch of these in `constants.h`

ljbade:
be worth doing some basic sanity checking too, e.g. tow range, sid ranges, iode/iodc etc

c/src/rtcm3_sbp_ephemeris.c
```diff
@@ -0,0 +1,140 @@
...
+      fit_interval = 98;
+    }
+  }
+
+  return fit_interval * 60 * 60;
+}
+
+void rtcm3_gps_eph_to_sbp(rtcm_msg_eph *msg_eph, msg_ephemeris_gps_t *sbp_gps_eph, struct rtcm3_sbp_state *state) {
+  /* RTCM gives wn module 1024, so take the current time and mask the lower 10 bits */
+  sbp_gps_eph->common.toe.wn = (state->time_from_rover_obs.wn & 0xFC00) + msg_eph->wn;
```
nsirola: This fails right after a week rollover when time is already on the next cycle but transmitted ephemeris still has toe on the previous cycle. (sure would be nice to be able to submodule LNSP's common part and use `gps_adjust_week_cycle` from there)

PR #99, opened 2018-08-07
Bare-bones command line rtcm to sbp converter.

martin-swift:


martin-swift:
@mbavaro - can you try this and make sure it reads all the available data in real-time?  It should...

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+/* Write the SBP packet to STDOUT. In theory, I could use
+   sbp_send_message(). */
+static void cb_rtcm_to_sbp(uint16_t msg_id, uint8_t length, uint8_t *buffer,
+                           uint16_t sender_id)
+{
+  /* SBP specifies little endian; this code should work on all hosts */
+  uint8_t tmpbuf[6];
+  tmpbuf[0] = SBP_PREAMBLE;
+  tmpbuf[1] = (uint8_t)msg_id;
+  tmpbuf[2] = (uint8_t)(msg_id >> 8);
```
mbavaro: I seem to remember that this is endianness independent (that is, should always result in `msg_id/256`) right?

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+#define FIFO_SIZE         (1 << 14)
+#define BUFFER_SIZE       (FIFO_SIZE - RTCM3_MSG_OVERHEAD - RTCM3_MAX_MSG_LEN)
+#define SBP_PREAMBLE      0x55
+
+static struct rtcm3_sbp_state state;
+
+/* Write the SBP packet to STDOUT. In theory, I could use
+   sbp_send_message(). */
+static void cb_rtcm_to_sbp(uint16_t msg_id, uint8_t length, uint8_t *buffer,
+                           uint16_t sender_id)
```
mbavaro: can we use `const` where possible? I know C does no enforcement, but sometimes it helps catching errors and readability..

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+  uint8_t tmpbuf[6];
+  tmpbuf[0] = SBP_PREAMBLE;
+  tmpbuf[1] = (uint8_t)msg_id;
+  tmpbuf[2] = (uint8_t)(msg_id >> 8);
+  tmpbuf[3] = (uint8_t)sender_id;
+  tmpbuf[4] = (uint8_t)(sender_id >> 8);
+  tmpbuf[5] = length;
+  ssize_t numwritten = write(STDOUT_FILENO, tmpbuf, sizeof(tmpbuf));
+  if (numwritten < (ssize_t)sizeof(tmpbuf)) {
+    fprintf(stderr, "Write failure at %d, %s\n", __LINE__, __FILE__);
```
mbavaro: what about `return` here and in general upon write failures?

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+}
+
+static void cb_base_obs_invalid(double timediff)
+{
+  fprintf(stderr, "Invalid base observation! timediff: %lf\n", timediff);
+}
+
+static uint16_t extract_msg_len(uint8_t *buf)
+{
+  return (((uint16_t)buf[1] << 8) | buf[2]) & 0x3FF;
```
mbavaro: I like this as it's faster, but did you consider using `getbitu()`?

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+  uint8_t crcbuf[2];
+  crcbuf[0] = (uint8_t)crc;
+  crcbuf[1] = (uint8_t)(crc >> 8);
+  numwritten = write(STDOUT_FILENO, crcbuf, sizeof(crcbuf));
+  if (numwritten < (ssize_t)sizeof(crcbuf)) {
+    fprintf(stderr, "Write failure at %d, %s\n", __LINE__, __FILE__);
+  }
+}
+
+static void cb_base_obs_invalid(double timediff)
```
mbavaro: `const` ?

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+   your actual message */
+static bool verify_crc(uint8_t *buf, uint16_t buf_len) {
+  uint16_t msg_len = extract_msg_len(buf);
+  assert(buf_len >= msg_len + RTCM3_MSG_OVERHEAD);
+  uint32_t computed_crc = crc24q(buf, 3 + msg_len, 0);
+  uint32_t frame_crc = (buf[msg_len + 3] << 16) |
+                       (buf[msg_len + 4] << 8) |
+                       (buf[msg_len + 5] << 0);
+  if (frame_crc != computed_crc) {
+    fprintf(stderr, "CRC failure! frame: %d computed: %d\n", frame_crc,
```
mbavaro: I suppose it wouldn't help printing the CRC as hexadecimal..

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+  assert(FIFO_SIZE > RTCM3_MSG_OVERHEAD + RTCM3_MAX_MSG_LEN);
+
+  /* set time from systime, account for UTC<->GPS leap second difference */
+  time_t ct_utc_unix = time(NULL);       
+  gps_time_t noleapsec = time2gps_t(ct_utc_unix); 
+  ct_utc_unix += get_gps_utc_offset(&noleapsec, NULL);
+  gps_time_t withleapsec = time2gps_t(ct_utc_unix);
+  gps_time_sec_t current_time;
+  current_time.tow = withleapsec.tow;
+  current_time.wn = withleapsec.wn;
```
mbavaro: nice!

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+
+  rtcm2sbp_init(&state, cb_rtcm_to_sbp, cb_base_obs_invalid);
+  rtcm2sbp_set_gps_time(&current_time, &state);
```
mbavaro: I cannot find these two 😢

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+  // rtcm2sbp_set_leap_second(18, &state); /* todo: worry about GLONASS later */
+  
+  uint8_t fifo_buf[FIFO_SIZE] = {0};
+  fifo_t fifo;
+  fifo_init(&fifo, fifo_buf, sizeof(fifo_buf));
+
+  uint8_t inbuf[BUFFER_SIZE];
+  ssize_t numread;
+  while ((numread = read(STDIN_FILENO, inbuf, BUFFER_SIZE)) > 0) {
```
mbavaro: we might need to do a blocking read of just 1 byte at a time for minimum RTCM -> SBP latency

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+        break;
+      }
+
+      uint8_t *rtcm_msg = &buf[index];
+      if (!verify_crc(rtcm_msg, FIFO_SIZE - index)) {
+        index++;
+        continue;
+      }
+      
+      rtcm2sbp_decode_frame(rtcm_msg, msg_len + RTCM3_MSG_OVERHEAD, &state);
```
mbavaro: this one is also somewhere else...

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+  fifo_t fifo;
+  fifo_init(&fifo, fifo_buf, sizeof(fifo_buf));
+
+  uint8_t inbuf[BUFFER_SIZE];
+  ssize_t numread;
+  while ((numread = read(STDIN_FILENO, inbuf, BUFFER_SIZE)) > 0) {
+    ssize_t numwritten = fifo_write(&fifo, inbuf, numread);
+    assert(numwritten == numread);
+
+    uint32_t index = 0;
```
mbavaro: I'd move this just before the `while( )` below?

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+/* This is a stand-alone tool that takes RTCM3 on stdin and writes
+   Swift Binary Protocol (SBP) on stdout.  It is designed to perform
+   well in embedded and cloud environments; likewise for live
+   vs. pre-recorded data.  Note that by default it sets the time to
+   the current system time, which may not be suitable for pre-recorded
+   data.  */
```
mbavaro: 👍  for the description, tidy!

README.md
```diff
@@ -0,0 +1,32 @@
...
+RTCM to SBP, a tool written in C for converting RTCM to SBP, and a
+variety of haskell tools that will convert between RTCM, SBP, and
+json.
+
+To build the C tool locally, first clone the libsbp repository into
+the c/ directory:
+
+```
+cd c
+git clone git@github.com:swift-nav/libsbp.git
```
benjamin0: Is there a way to bring in this dependency more formally?

c/src/localmods/edc_local.h
```diff
@@ -0,0 +1,18 @@
+/*
```
benjamin0: Can we add the standard Swift header here?

c/src/lsncopied/README.md
```diff
@@ -0,0 +1,6 @@
+These files have been copied from the refactor directory of
+libswiftnav-private; the only modifications are those required for
+compilation.
+
+Again, this is intended to be temporary until we have an effective
+utils library.
```
benjamin0: Can you make an issue for this task?

README.md
```diff
@@ -0,0 +1,32 @@
...
+
+Then follow the usual steps:
+
+```
+mkdir build
+cd build
+cmake ..
+make -j8
+```
+
```
benjamin0: Can you add an example or two of how to run the program(s)?

c/src/lsncopied/almanac.h
```diff
@@ -0,0 +1,183 @@
...
+#define GPS_LNAV_ALM_SF_W C_1_2P23
+/** IS-GPS-200H Table 20.6: 2^-23 */
+#define GPS_LNAV_ALM_SF_M0 C_1_2P23
+/** IS-GPS-200H Table 20.6: 2^-20 */
+#define GPS_LNAV_ALM_SF_AF0 C_1_2P20
+/** IS-GPS-200H Table 20.6: 2^-38 */
+#define GPS_LNAV_ALM_SF_AF1 C_1_2P38
+
+/** Structure containing the GPS almanac for one satellite. */
+typedef struct {
```
benjamin0: Is any of this stuff from here down actually used anywhere in this repo?

README.md
martin-swift: sure - we can add libsbp as a submodule; that is exactly what kicked off my discussion with @anth-cole . I was under the impression that was inadvisable since buildroot could potentially pull in a different version of libsbp.

c/src/lsncopied/almanac.h
martin-swift: I don't think so; I can chop it if you want.

c/src/rtcm3tosbp.c
martin-swift: thanks!

c/src/rtcm3tosbp.c
martin-swift: Can you elaborate?  I am unclear what you are trying to convey...

c/src/rtcm3tosbp.c
martin-swift: That was the part I wanted to fix the most :) Note that fread() is subtly different from read(); by going one layer of abstraction closer to the OS, we get some magical benefits for free - namely read() will return all the bytes it has available, or block if nothing is available. So we should be able to read from large files with good performance as well as have real-time observations (and the process should block when there is no data).  This is exactly the part I want you to test, since you understand the timing requirements for this data far better than I do.  Please tell me - is there any latency issue with this code?

c/src/rtcm3tosbp.c
martin-swift: src/rtcm3_sbp.c

c/src/rtcm3tosbp.c
martin-swift: src/rtcm3_sbp.c

c/src/rtcm3tosbp.c
martin-swift: Now that I think about it, exit() is probably what we want.  If this happens, something is unrecoverably bad and it's better to find out sooner and fix it.

README.md
mfine: libsbp is submoduled in on master now.

c/include/rtcm3_sbp.h
```diff
@@ -39,6 +39,10 @@
 #define SBP_GLO_FCN_OFFSET 8
 #define SBP_GLO_FCN_UNKNOWN 0
 
+#define RTCM3_PREAMBLE      0xD3
+#define RTCM3_MSG_OVERHEAD     6
+#define RTCM3_MAX_MSG_LEN   1023
```
mfine: MAX_MSG_LEN is 1024.

c/src/rtcm3tosbp.c
martin-swift: This is defined in rtcm3_sbp.h and there are likely lost of other consumers that would be affected, so I don't want to touch that.

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+{
+  /* SBP specifies little endian; this code should work on all hosts */
+  uint8_t tmpbuf[6];
+  tmpbuf[0] = SBP_PREAMBLE;
+  tmpbuf[1] = (uint8_t)msg_id;
+  tmpbuf[2] = (uint8_t)(msg_id >> 8);
+  tmpbuf[3] = (uint8_t)sender_id;
+  tmpbuf[4] = (uint8_t)(sender_id >> 8);
+  tmpbuf[5] = length;
+  ssize_t numwritten = write(STDOUT_FILENO, tmpbuf, sizeof(tmpbuf));
```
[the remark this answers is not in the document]
martin-swift: Since I am using read(), I figured I would match that with write() since our output is binary anyway.  I am using stderr so I can format error messages with fprintf().

c/src/CMakeLists.txt
```diff
@@ -4,6 +4,11 @@ add_library(gnss_converters rtcm3_sbp.c)
 target_link_libraries(gnss_converters m sbp rtcm)
 target_include_directories(gnss_converters PUBLIC ${PROJECT_SOURCE_DIR}/include)
 target_include_directories(gnss_converters PUBLIC ${PROJECT_SOURCE_DIR}/src)
+target_include_directories(gnss_converters PUBLIC ${PROJECT_SOURCE_DIR}/localmods)
+target_include_directories(gnss_converters PUBLIC ${PROJECT_SOURCE_DIR}/lsncopied)
```
mfine: localmods and lsncopied is clunky - can we just bring the source and header files into the src directory?

c/src/localmods/edc_local.c
```diff
@@ -0,0 +1,163 @@
...
+  if (invert) {
+    acc ^= 0xFFu;
+  }
+  b = (acc >> shift) & 0xFFu;
+  crc = ((crc << 8) & 0xFFFFFFu) ^ crc24qtab[((crc >> 16) ^ b) & 0xFFu];
+
+  return crc;
+}
+
+uint16_t crc16_ccitt(const uint8_t *buf, uint32_t len, uint16_t crc) {
```
mfine: You should be able to pick this up from libsbp library.

c/src/rtcm3tosbp.c
```diff
@@ -0,0 +1,158 @@
...
+    fifo_size_t bytes_avail = fifo_peek(&fifo, buf, FIFO_SIZE);
+
+    while (index < (bytes_avail - RTCM3_MSG_OVERHEAD)) {
+      if (RTCM3_PREAMBLE != buf[index]) {
+        index++;
+        continue;
+      }
+
+      uint16_t msg_len = extract_msg_len(&buf[index]);
+      if ((msg_len == 0) || (msg_len > RTCM3_MAX_MSG_LEN)) {
```
mfine: msg_len should gets masked off with 0x3FF - shouldn't be possible to eclipse 1023.

c/src/rtcm3tosbp.c
mbavaro: I see, so `read()` will return immediately a short amount of data if less than `BUFFER_SIZE` was available in the pipe.. nice! There shouldn't be any latency like this.
I think `fread()` shouldn't be so bad either as it is buffered so it shouldn't perform any actual peripheral access when reading 1 Byte at a time, but this solution is surely more elegant!

c/src/rtcm3tosbp.c
mbavaro: fair enough 😄

c/src/lsncopied/README.md
martin-swift: https://swift-nav.atlassian.net/browse/MAP-509

README.md
martin-swift: done

c/include/rtcm3_sbp.h
martin-swift: So what's up with this in check_rtcm3.c:
```
    if (message_size > 1023) {
      /* too large message */
      buffer_index++;
      continue;
    }
```

c/src/CMakeLists.txt
martin-swift: we could, but the idea there is those directories should go away once we have the utils library.  I just wanted to be clear that that stuff is not really part of gnss-converters.

c/src/rtcm3tosbp.c
martin-swift: correct... do you think it is bad to have this extra sanity check? So is the max message length 1023 or 1024? if it's 1024, then I shouldn't be masking with 0x3FF....

c/include/rtcm3_sbp.h
martin-swift: Check page 263 of RTCM 10403.3.  I believe that max msg len is 1023; that is also consistent with it being a 10 bit field.

mfine: Ahh, I though you got the whole 1024 bits from the 10 bit length field - weird.

c/src/CMakeLists.txt
mfine: I think bringing them into the src directory is fine. There's not that many files - and there's already copied code in places. The extra directories are clunky.

c/src/rtcm3tosbp.c
mfine: Sounds like you could hit 1024 from the masked off bits.

c/include/rtcm3_sbp.h
mfine: So looking at their wording further in the document, I think this means 1024 length - because they use the term bytes 0 to 1023, where 0 is inclusive:

screen shot 2018-08-07 at 16 45 53

it's weird language.

martin-swift: I don't see anything odd about that... you get 1024 possible values, from 0 to 1023, inclusive.

c/src/rtcm3tosbp.c
martin-swift: I feel like we sorted this out on the other thread, but I have to disagree - b1111111111 == 1023, no?

c/include/rtcm3_sbp.h
mfine: Right, but they're using the same usage on page 263:

screen shot 2018-08-07 at 18 25 37

and `RTCM3_MAX_MSG_LEN` is saying that the length of a message can be 1023 - but the max message length is 1024 - the position of the last byte is 1023. The max value of the length field should check if it's `> 1024` (which is redundant with the 0x3ff masking. The check in check_rtcm3.c is wrong.

mfine: Ugh, I'm getting tripped up because 0 is a valid value - my bad.

mfine:
Still :-1: on the clunky `lsncopied` and `localmods` directories and their contents being wholesale vendored in. I'd like to just see the relevant functionality being brought in. This is a small tool - roughly a hundred lines - and its vendoring in 3K lines of code:

```
     174 localmods/fifo_local.c
      46 localmods/fifo_local.h
     183 lsncopied/almanac.h
      60 lsncopied/bits.h
      87 lsncopied/common.h
     654 lsncopied/constants.h
     124 lsncopied/edc_local.c
      27 lsncopied/edc_local.h
     864 lsncopied/gnss_time.c
     296 lsncopied/gnss_time.h
     594 lsncopied/signal.h
    3109 total
```

I can't recall anywhere where we do this kind of thing or use directories like `xxxcopied` or `localmods`. I'd just vendor in the needed functionality into the directory. And I don't agree with the ugly placeholder to be removed at a later data when everything is magically fixed - it leaves thing messy in the meantime, the more times than not things never get fixes, people never go back and update the code.

I'd name the executable and file `run_rtcm3_sbp` and `run_rtcm3_sbp.h`.

mfine:
If the current vendoring approach is kept, I think functionality should get moved out to its own repo.

martin-swift:
Might not be a bad idea... ok, working on it.  I was trying to figure out exactly what target_include_directories does; if that's just for local compilation, mixing it all together is probably fine.  But if we are OK with another repository, sure, I'll cook one up and link against that until we have the real utils library. Nice and clean.  I like it.  Also, might have a little CRC problem that @mbavaro identified; will squash that too.

martin-swift:
Also, why do you advocate run_rtcm3_sbp?  None of the haskell tools are named like that.  It is also not in keeping with the Unix naming convention, and in my opinion, substantially less descriptive.

mfine:
The `run_*` is a convention that's been used in lsn:

```
libswiftnav-private/test_pvt_engine/orion/run_gage_ionosphere_model.cc
libswiftnav-private/test_pvt_engine/orion/run_atmospheric_model.cc
libswiftnav-private/test_pvt_engine/orion/run_atmospheric_model_utils.cc
libswiftnav-private/test_pvt_engine/replay/run_vehicle_dynamics_filter.cc
libswiftnav-private/test_pvt_engine/replay/run_network_filter.cc
libswiftnav-private/test_pvt_engine/replay/run_filter.cc
libswiftnav-private/test_pvt_engine/replay/run_filter_utils.cc
```

Don't think it's hard fast, but seems to be a convention when giving "exec" functionality to something. The naming is all over the place though, in general.

martin-swift:
Ok, take a look, especially @mfine ; This should build on travis just fine once I either make libutils-tmp public or sort out this key business.

martin-swift:
Ok, as soon as @anth-cole makes libswfitnav public, I think this is finally in good merge-able shape.

martin-swift:
Note also that gnss-converters now gets built as a shared library.

PR #106, opened 2018-08-15
Add Galileo and Beidou SSR corrections support

scarcanague:
I ran clang-format (in separate commits), so probably better to review each commit separately

### DIFF
--- a/c/src/rtcm3_sbp_internal.h
+++ b/c/src/rtcm3_sbp_internal.h
@@ -225,6 +225,4 @@ void rtcm3_ssr_orbit_clock_to_sbp(rtcm_msg_orbit_clock *msg_orbit_clock, struct 
 void rtcm3_ssr_code_bias_to_sbp(rtcm_msg_code_bias *msg_code_biases, struct rtcm3_sbp_state *state);
 void rtcm3_ssr_phase_bias_to_sbp(rtcm_msg_phase_bias *msg_phase_biases, struct rtcm3_sbp_state *state);
 
-code_t constellation_to_l1_code(const enum constellation_e constellation);
-
 #endif /* GNSS_CONVERTERS_RTCM3_SBP_H */
--- a/c/src/rtcm3_sbp_internal.h
+++ b/c/src/rtcm3_sbp_internal.h
@@ -223,4 +223,6 @@ void rtcm3_ssr_orbit_clock_to_sbp(rtcm_msg_orbit_clock *msg_orbit_clock, struct 
 void rtcm3_ssr_code_bias_to_sbp(rtcm_msg_code_bias *msg_code_biases, struct rtcm3_sbp_state *state);
 void rtcm3_ssr_phase_bias_to_sbp(rtcm_msg_phase_bias *msg_phase_biases, struct rtcm3_sbp_state *state);
 
+code_t constellation_to_l1_code(const enum constellation_e constellation);
+
 #endif /* GNSS_CONVERTERS_RTCM3_SBP_H */
--- a/c/src/rtcm3_sbp_internal.h
+++ b/c/src/rtcm3_sbp_internal.h
@@ -178,6 +178,8 @@ void compute_glo_time(u32 tod_ms,
                       const gps_time_sec_t *rover_time,
                       struct rtcm3_sbp_state *state);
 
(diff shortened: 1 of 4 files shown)


c/src/rtcm3_sbp_internal.h
```diff
@@ -223,4 +223,6 @@ void rtcm3_ssr_orbit_clock_to_sbp(rtcm_msg_orbit_clock *msg_orbit_clock, struct
 void rtcm3_ssr_code_bias_to_sbp(rtcm_msg_code_bias *msg_code_biases, struct rtcm3_sbp_state *state);
 void rtcm3_ssr_phase_bias_to_sbp(rtcm_msg_phase_bias *msg_phase_biases, struct rtcm3_sbp_state *state);
 
+code_t constellation_to_l1_code(const enum constellation_e constellation);
```
anth-cole: since we now include parts of `libswiftnav/common` in this, it's good to prefix these global functions with a `static` or define them in the header as a `static inline`

anth-cole: if this function is only needed in the SSR module, then better to just declare it `static` in the `.c` file and not declare it in the header at all - this limits the scope to only this module

c/src/rtcm3_sbp_ssr.c
```diff
@@ -1,30 +1,66 @@
...
-      compute_glo_time(msg_orbit_clock->header.epoch_time * S_TO_MS, &sbp_orbit_clock->time, &state->time_from_rover_obs, state);
-      sbp_orbit_clock->sid.code = 3;
+    if (msg_orbit_clock->header.constellation == CONSTELLATION_GLO) {
+      compute_glo_time(msg_orbit_clock->header.epoch_time * S_TO_MS,
+                       &sbp_orbit_clock->time,
+                       &state->time_from_rover_obs,
+                       state);
+    } else {
+      // GPS / GAL / BDS / QZSS / SBAS use GPS-like type of time
+      compute_gps_message_time(msg_orbit_clock->header.epoch_time * S_TO_MS,
```
anth-cole: Are you sure this is right? I thought they used their system time and we'd need to account for the different offsets (like BDS uses a different leap second to GPS time?)

c/src/rtcm3_sbp_ssr.c
```diff
@@ -1,30 +1,66 @@
...
+                               &sbp_orbit_clock->time,
+                               &state->time_from_rover_obs);
     }
-    if(!gps_time_valid(&sbp_orbit_clock->time)){
+    sbp_orbit_clock->sid.code =
```
anth-cole: was I missing this? nit: move it below the time valid check

c/tests/check_rtcm3_ssr.c
```diff
@@ -27,17 +27,25 @@ static bool gps_code_bias_processed = false;
 static bool gps_phase_bias_processed = false;
 static bool glo_orbit_clock_processed = false;
 static bool glo_code_bias_processed = false;
+static bool gal_orbit_clock_processed = false;
```
anth-cole: I left it off as we didn't have data, but is it worth adding a glo phase bias test too if that's supported now?

c/src/rtcm3_sbp_internal.h
scarcanague: Done. It was used in the test, but I think it is better as `static inline` indeed, so hardcoded the L1 values in the test

c/src/rtcm3_sbp_ssr.c
scarcanague: Yes, realized Beidou had a fixed number of leap seconds to GPS time. Corrected.

c/tests/check_rtcm3_ssr.c
scarcanague: CNES does not provide phase biases for GLONASS (only orbit/clock and code biases).

c/src/rtcm3_sbp_ssr.c
scarcanague: It was not missing, it was above in each `if`:
https://github.com/swift-nav/gnss-converters/pull/106/files#diff-a721b0b5310d9e081c51beee1afd9d14L22
Good point, I moved it down

PR #110, opened 2018-08-22
Add multi constellation eph support

scarcanague:
cc @martin-swift


martin-swift:
Unless there is a compelling reason to keep them around, can we start deleting branches after merge?

PR #111, opened 2018-08-22
Fix station id mask

mfine:
Make it so the high bit in the sender id will always get set. Similar to haskell converters.

/cc @benjamin0 @scarcanague


scarcanague:
LGTM!

PR #119, opened 2018-09-11
DO NOT MERGE - yet. librtcm is pointed at devel branch.

martin-swift:


martin-swift:
@mfine is it easy to check if remove EXCLUDE_FROM_ALL breaks your compile again?  If not, I will do it, but I don't quite have orion compiling yet.

martin-swift:
@mfine I verified that after my fixes to libswiftnav that removed the BUILD_SHARED, you can build all targets here without interfering with the orion compile.  However, when you update to this version of gnss-converters, you will have to change one include.

PR #120, opened 2018-09-14
EX-341 RTCM encoder

nsirola:
The basics of RTCM encoder and `sbp2rtcm` functionality. 

Pulls in https://github.com/swift-nav/librtcm/pull/61

See minimal usage example in `sbp2rtcm`: https://github.com/swift-nav/gnss-converters/pull/120/files#diff-01212e3190c3b8042cece5ccee1fd8fcR64

## TODO
- maybe split the `rtcm3_sbp.c` into three files, e.g.: `rtcm3_to_sbp.c`, `sbp_to_rtcm3.c`, and `rtcm3_common.c`?

Note that MSM conversion will be done separately under EX-351

## Known issues

- SBP sender ID <-> RTCM station ID conversion is not reversible as SBP is 16bit and RTCm is 12bit, and on top of that bit 7 is always set on in RTCM->SBP conversion
- lock time conversion is not reversible, e.g. SBP lock indicator 15 (lock time >= 524.288 s) is converted to RTCM lock indicator 106 (lock time >= 520 s), which when converted back to SBP maps to indicator value 14 (262.144-524.288 s). 

## Open questions
- is there need to configure the receiver specific flags, e.g. quarter-cycle conversion

## Testing
Converted a 18 hour live data log first with `sbp2rtcm` and then back with `rtcm2sbp`  and compared the converted SBP to the original. The log already had a fair amount of missing OBS messages so also broken obs sequence handling got covered.
- all the GPS  L1 observations with carrier phase flag came through
- after the first 20 seconds, all the GLO L1 observations with carrier phase flag came through (first GLO ephemerides needed for FCNs)
- observations without valid carrier are not converted (intended)
- L2 observations are converted only if the corresponding L1 is (limitation of RTCM 1004/1012 format)
- about 4% of GPS pseudoranges had rounding jitter of 2cm (probably can't be helped)
- GLO pseudoranges were always equal (they have 1 bit more range in RTCM than GPS ones)
- positive carrier phase integer parts were always equivalent (to be investigated why there were negative carrier phases in the SBP!)
- fractional carrier phases were always equivalent
- CN0 were always equivalent
- (Doppler is not included in 1004/1012)
- lock time indicator was mostly off by 1 (see known issues above)
- week number was off (to be fixed by https://github.com/swift-nav/gnss-converters/pull/134)


ljbade:
Something we will need to carefully check is the I/Q carrier phase alignment from Piksi in SBP. RTCM MSM has very specific rules about the alignment of the different signal phases. And for the legacy RTCM messages they also have specific rules which are different from MSM.

nsirola:
@ljbade I've understood Piksi corrects for quarter-cycle shifts internally so that they are also aligned in the SBP output. 

RTCM message has a flag to indicate if quarter-cycle shifts have been removed or not, but on the other hand, it also says that "Manufacturers not listed in [Table 3.5-7] shall not apply 1/4 cycle shifts"  ...

ljbade:
@nsirola so Piksi already aligns the L2C obs to L2P?

Whats confusing is basically:
RTCM 3.0 legacy messages - "Manufacturers not listed in [Table 3.5-7] shall not apply 1/4 cycle shifts"
RTCM 3.2 MSM - "The phase observations shall be shifted by the respective fraction of a cycle, either directly by the receiver or by a correction program or the MSM conversion program, prior to MSM stream generation, to align them to the reference signal. The reference signal shall be selected according to Table 3.5-73. The reference signal may or may not be tracked." - and for GPS L2 the reference signal is L2P

So you can see they are contradictory, so depending on what we get from Piksi either it will be correct for legacy or for MSM. So we will need to apply this shift somewhere.

nsirola:
@ljbade @mbavaro We do some phase adjusting in FW, but not comprehensively: https://github.com/swift-nav/piksi_firmware_private/blob/master/src/acq/manage.c#L1059-L1063

I think I'll put the quarter cycle flag to the undefined (00) value for now. 

At least we know that the SBP output cannot change depending on which RTCM out mode is selected, so I think the plan forward is:
- SBP output should have the carrier phases aligned according to RTCM specification (Table 3.5-73)
- the MSM output (when implemented) can then use SBP carrier values directly
- when encoding legacy RTCM 1004/1012 messages, de-align the carrier phases
- add a function in `libswiftnav` for phase alignment according to RTCM Table 3.5-73, and use it both for alignment in FW and de-alignment in legacy RTCM encoding. (Note that FW currently seems to do the opposite for GAL signals at https://github.com/swift-nav/piksi_firmware_private/blob/master/src/acq/manage.c#L1059-L1063)

nsirola:
Opened an issue on alignment in FW: https://swift-nav.atlassian.net/browse/EX-361

c/tests/check_rtcm3.c
```diff
@@ -544,8 +539,88 @@ void test_RTCM3(const char *filename,
...
+
+  sbp_state_set_io_context(&s, fp);
+
+  while (!feof(fp)) {
+    sbp_process(&s, &sbp_read_file);
+  }
+  fclose(fp);
+}
+
+void rtcm_sanity_check_cb(u8 *buffer, u8 length, void *context) {
```
pmiettinen: static?

c/tests/check_rtcm3.c
```diff
@@ -544,8 +539,88 @@ void test_RTCM3(const char *filename,
...
+  msg_ephemeris_glo_t *e = (msg_ephemeris_glo_t *)msg;
+
+  /* extract just the FCN field */
+  sbp2rtcm_set_glo_fcn(e->common.sid, e->fcn, &out_state);
+}
+
+void test_SBP(const char *filename,
+              void (*cb_sbp_to_rtcm)(u8 *buffer, u8 length, void *context),
+              gps_time_sec_t current_time_) {
+  (void)current_time_;
```
pmiettinen: This seems closer to full e2e application test than unit test. Would it be more suitable to execute the actual binary with `system()` or `execvp()` and compare the output to a verified file?

c/tests/check_rtcm3.c
```diff
@@ -1085,6 +1188,7 @@ Suite *rtcm3_suite(void) {
   tcase_add_test(tc_gpp_biases, test_bias_gpp_sok);
   tcase_add_test(tc_gpp_biases, test_bias_gpp_tps1);
   tcase_add_test(tc_gpp_biases, test_bias_gpp_trm);
+  // tcase_add_test(tc_gpp_biases, test_sbp_to_rtcm_legacy);
```
pmiettinen: Remove?

c/src/rtcm3_sbp.c
```diff
@@ -1849,3 +1956,278 @@ void rtcm3_msm_to_sbp(const rtcm_msm_message *msg,
...
+  }
+
+  assert(glo_tod_s >= 0 && glo_tod_s < (SEC_IN_DAY + 1));
+  return glo_tod_s;
+}
+
+static void sbp_buffer_to_rtcm3(struct rtcm3_out_state *state) {
+  u16 station_id = sbp_2_rtcm_sender_id(state->sender_id);
+
+  rtcm_obs_message gps_obs;
```
pmiettinen: Could add a helper `rtcm_init_obs(rtcm_obs_message &gps_obs, constellation_t c)`.

c/src/rtcm3_sbp.c
```diff
@@ -1849,3 +1956,278 @@ void rtcm3_msm_to_sbp(const rtcm_msm_message *msg,
...
+  glo_obs.header.smooth = PIKSI_SMOOTHING_INTERVAL;
+
+  u8 n_gps = 0;
+  u8 n_glo = 0;
+
+  /* NOTE: assumes that L1 obs are sorted before the L2 obs */
+
+  for (u8 i = 0; i < state->n_sbp_obs; i++) {
+    packed_obs_content_t *sbp_obs = &(state->sbp_obs_buffer[i]);
+    if (CODE_GPS_L1CA == sbp_obs->sid.code ||
```
pmiettinen: Loop seems a handful, additional helper could help to tone down the function length but seems questionable is it worth it.

c/tests/check_rtcm3.c
nsirola: Removed

c/src/rtcm3_sbp.c
nsirola: Added

c/src/rtcm3_sbp.c
nsirola: Reformatted it from `if`-`else if` monstrosity into a `switch` tree and extracted one lookup helper function.

c/src/rtcm3_sbp.c
```diff
@@ -421,6 +461,82 @@ static bool is_msm_active(const gps_time_sec_t *current_time,
...
+  }
+  return 0;
+}
+
+/* Encode RTCM frame into the given buffer, returns frame length */
+u16 encode_rtcm3_frame(const void *rtcm_msg,
+                       u16 message_type,
+                       u8 *frame,
+                       struct rtcm3_out_state *state) {
+  u16 byte = 3;
```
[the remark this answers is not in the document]
nsirola: It would... but this consistent with the notation used in the corresponding decode function.

c/tests/check_rtcm3.c
nsirola: That would be interesting to set up

PR #121, opened 2018-09-15
Fix link against sbp-static

mbavaro:
This seems to fix the build when CMake is invoked with 
```
cmake ../ -DCMAKE_SYSTEM_NAME=Linux -DBUILD_SHARED_LIBS=OFF
```


c/src/CMakeLists.txt
```diff
@@ -5,7 +5,12 @@ set(gnss_converters_HEADERS
   )
 
 add_library(gnss_converters rtcm3_sbp.c rtcm3_sbp_ephemeris.c rtcm3_sbp_ssr.c sbp_nmea.c nmea.c)
+
+if (BUILD_SHARED_LIBS)
 target_link_libraries(gnss_converters m swiftnav sbp rtcm)
+else (BUILD_SHARED_LIBS)
```
martin-swift: I like else()  and endif().  I don't even know why CMake syntax allows this; to me it's a red herring.

martin-swift:
Can you post the error you are seeing?  I have never seen a link problem before, but I have also never explicitly turned off shared libs...

c/src/CMakeLists.txt
mbavaro: I don't like them either, but actually my CMake complained about a syntax error without these! 😊

mbavaro:
```
cmake ../ -DBUILD_SHARED_LIBS=OFF
```

```Scanning dependencies of target rtcm3tosbp
[ 97%] Building C object src/CMakeFiles/rtcm3tosbp.dir/rtcm3tosbp.c.o
[100%] Linking C executable rtcm3tosbp
/usr/bin/ld: cannot find -lsbp
collect2: error: ld returned 1 exit status
src/CMakeFiles/rtcm3tosbp.dir/build.make:97: recipe for target 'src/rtcm3tosbp' failed
make[2]: *** [src/rtcm3tosbp] Error 1
CMakeFiles/Makefile2:423: recipe for target 'src/CMakeFiles/rtcm3tosbp.dir/all' failed
make[1]: *** [src/CMakeFiles/rtcm3tosbp.dir/all] Error 2
Makefile:127: recipe for target 'all' failed
make: *** [all] Error 2
```

c/src/CMakeLists.txt
mbavaro: don't remember.. but happy to move to `()` if it's better.. I'll update the PR

martin-swift:
So.... can I propose an idea?  First, can you make a pull that updates libsbp to master?  I think there is a compile error against a unit test (last time I checked, at least).  Then can we just fix the CMake in libsbp?  I am happy to do that now if you want.  (could use a break from this HD5 stuff anyway)

martin-swift:
ok, if https://github.com/swift-nav/libsbp/pull/594 is approved, we can stop checking this shared libs flag altogether.

mbavaro:
obsoleted

PR #128, opened 2018-09-24
[MAP-629] Fix NMEA when DR is enabled

anth-cole:
Accept SBP messages even if marked invalid and rely on time matching to send nmea messages in order to support DR mode. This alos removes a lot of dependencies on the gps time message and instead relys on the utc time message. Also add NMEA dead reckoning output mode E

### DIFF
--- a/c/include/gnss-converters/nmea.h
+++ b/c/include/gnss-converters/nmea.h
@@ -30,6 +30,7 @@
 #define NMEA_GGA_QI_EST 6
 #define NMEA_GGA_QI_MANUAL 7
 #define NMEA_GGA_QI_SIM 8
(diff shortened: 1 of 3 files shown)


anth-cole:
removing the `sbp_msg_time` actually came from the fact that the internal message behavior has changed. Previously, the gps time message was unpopulated when marked invalid, now it's still populated (along with the utc time message) this means I can now get all the time information needed from the utc message and I can remove the dependency on the gps time message for some of the messages.

benjaminaltieri:
Ah ok, yeah this is definitely not my strong area. The code looks good. If not in this fix PR, we should at least add a ticket to circle back and add test cases if applicable.

PR #134, opened 2018-09-26
Add WN adjusting to eph decoders

mbavaro:
Right now the converter does not output Glonass because the leap seconds are missing. 
3rd party MEs with RTCM output do have Glonass support, so try to work around the uncertainty of absolute time by adding explicit week decoding from the RTCM ephemerides messages. 
The SBP version of the WN is used
https://github.com/swift-nav/libsbp/blob/master/c/include/libsbp/observation.h#L138
so that hopefully we are comparing always with unrolled values aligned to the GPS system time.

TODO: maybe add explicit parsing of RTCM1013 and DF054 ?

image


c/include/gnss-converters/rtcm3_sbp.h
```diff
@@ -54,6 +55,8 @@ typedef enum {
 
 struct rtcm3_sbp_state {
   gps_time_sec_t time_from_rover_obs;
+  utc_params_t utc_params;
+  bool utc_params_known;
```
mbavaro: right now this is never set, but imagine if some RTCM message carries this info at some point..

nsirola: I'm not sure if it servers any purpose to add these fields at this point as there are no RTCM or SBP messages with this payload. Any code handling these parameters would be just dead weight.

c/src/rtcm3_sbp.c
```diff
@@ -972,6 +979,21 @@ void rtcm2sbp_set_leap_second(s8 leap_seconds, struct rtcm3_sbp_state *state) {
...
+  if (abs(gpst_sec.wn - wn_ref) > 1) {
+    /* this only corrects big discrepancies */
+    gpst_sec.wn = wn_ref;
+  }
+  utc_params_t *p = NULL;
+  if (state->utc_params_known) {
+    p = &(state->utc_params);
+  }
+  gps_time_t gt = {.wn = gpst_sec.wn, .tow = gpst_sec.tow};
+  double gps_utc_offset = get_gps_utc_offset(&gt, p);
```
nsirola: For the time being, this could just as well be written `get_gps_utc_offset(&gt, NULL)` with a comment stating that it will use the hard-coded leap second table from `libswiftnav`, to be updated in software as new leap seconds are announced.

c/src/rtcm3_sbp.c
```diff
@@ -972,6 +979,21 @@ void rtcm2sbp_set_leap_second(s8 leap_seconds, struct rtcm3_sbp_state *state) {
...
+  rtcm2sbp_set_leap_second(gps_utc_offset, state);
```
nsirola: Safer to use `round(gps_utc_offset)`, otherwise the function will truncate it to integer.

c/src/rtcm3_sbp.c
```diff
@@ -69,7 +72,7 @@ void rtcm2sbp_init(struct rtcm3_sbp_state *state,
   rtcm_init_logging(&rtcm_log_callback_fn, state);
 }
 
-static void normalize_gps_time(gps_time_sec_t *t) {
+static void normalize_gps_sec_time(gps_time_sec_t *t) {
```
nsirola: Good catch! 
Although I wonder if we could now use `gps_time_t` from `libswiftnav` throughout and benefit from the existing utility functions for it instead of this integer version from `libsbp` that needs to have all helper functions re-implemented here.

c/src/rtcm3_sbp.c
nsirola: It's still better than hard-coding 18 here, or even adding a separate define for the leap second offset.

mbavaro:
@nsirola now?

mbavaro:
There shouldn't be any difference.. I like `rint()` as it is more balanced and generally available.. matter of taste I guess?

mbavaro:
Merging, give me a shout in case the sky falls.

PR #135, opened 2018-10-01
EX-363: Fix the 1230 indicator flag handling

nsirola:
Emit zero corrections in SBP GLO BIASES message if the base reports in 1230 that GLO pseudorange and carrier are already aligned. (see discussion in https://github.com/swift-nav/gnss-converters/pull/57#discussion_r221235497)


c/src/rtcm3_sbp.c
```diff
@@ -937,17 +937,26 @@ void rtcm3_1033_to_sbp(const rtcm_msg_1033 *rtcm_1033,
...
+        round(rtcm_1230->L1_CA_cpb_meter * GLO_BIAS_RESOLUTION);
+    sbp_glo_bias->l1p_bias =
+        round(rtcm_1230->L1_P_cpb_meter * GLO_BIAS_RESOLUTION);
+    sbp_glo_bias->l2ca_bias =
+        round(rtcm_1230->L2_CA_cpb_meter * GLO_BIAS_RESOLUTION);
+    sbp_glo_bias->l2p_bias =
+        round(rtcm_1230->L2_P_cpb_meter * GLO_BIAS_RESOLUTION);
+  } else {
+    /* no biases sent */
+    sbp_glo_bias->mask = 0;
```
anth-cole: Rather than this - I think instead we should send an SBP message 1230 in with mask set `sbp_glo_bias->mask = rtcm_1230->fdma_signal_mask;` and with the bias values set to zero - The reason being as you pointed out is the obs are valid, but without the receipt of the SBP glo bias message we won't try and fix glo ambiguities.

nsirola: Ok, changed the logic. Will review the corresponding logic in FW side tomorrow and test on bench.

nsirola:
Buildroot test PR at https://github.com/swift-nav/piksi_buildroot/pull/886

nsirola:
I was trying to find test cases to verity that this improves results, but on every RTCM stream and logs I have available, if the indicator is 1, the biases are already all zero, so the negative sign bug probably never manifested itself. Merging this in any case so that the code is correct even if it is only of cosmetic value.

PR #136, opened 2018-10-01
Update libsbp pointer

mbavaro:
self explanatory? solves the static build script


nsirola:
At some point we should point to a tagged release in `libsbp` and update buildroot to point to the same. Libsbp master branch already has e.g. interface changes to the `read` and `write` functions that require several changes in buildroot.

nsirola:
The commit for tagged v2.4.1 is 58442f9b303cfde659feab976ec8a1a0958ae9a2

mbavaro:
thanks Niilo

PR #138, opened 2018-10-05
EX-370: Skip SBP observations with RAIM flag in sbp2rtcm

nsirola:
Do not encode RAIM-flagged observations into the RTCM out stream.


nsirola:
Closing in favor of https://github.com/swift-nav/gnss-converters/pull/143

PR #140, opened 2018-10-08
Fix for [MAP-664]

anth-cole:
When the ns field of the UTC message is 999999999 the time stamp is mislabeled to the previous second and marked as 0.1, where the ns field should roll over into the second field.

So a UTC time of 170831.99999999 was being marked as 170831.1 and not 170832.0

@nsirola can you check and see if this makes sense to you? I think we've seen this before in other cases - is there a better fix?

### DIFF
--- a/c/src/nmea.c
+++ b/c/src/nmea.c
@@ -193,15 +193,17 @@ static void get_utc_time_string(bool time,
 
   if (time) {
     /* Time (UTC) */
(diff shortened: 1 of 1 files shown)


nsirola:
This fails with e.g. 170859.999999, rounding it to 170860.0. I'll look up how we solved this previously (and add some unit tests while I'm at it).

nsirola:
Improved solution at https://github.com/swift-nav/gnss-converters/pull/141

PR #143, opened 2018-10-10
EX-351: SBP to MSM conversion

nsirola:
Pulls in `librtcm` PR https://github.com/swift-nav/librtcm/pull/66

Add conversion from SBP observation message sequence into a set of MSM4 or MSM5 messages. (note that these are accurate enough for lossless reproduction of SBP observations, no need for the high-accuracy MSM6/7)

Add unit tests for the MSM utility functions, as well as SBP->MSM->SBP round trip conversion for a single observation set.

## Interface

Add a basic interface for toggling the sending of legacy and/or MSM observations and their type:
```
void sbp2rtcm_set_rtcm_out_mode(msm_enum value, struct rtcm3_out_state *state);
```
Valid values for `msm_enum` are `MSM4`, `MSM5`, and `MSM_UNKNOWN` which will result in sending legacy 1004/1012 obs.

## Testing

Buildroot test PR at https://github.com/swift-nav/piksi_buildroot/pull/926

Verified with a desktop master-rover Piksi pair that every observation logged in the base station's SBP log is identically reproduced in the rover's SBP log as a base observation.

Overnight test run on bench zero-baseline Piksi-to-Piksi setup and no issues found.


nsirola:
@anth-cole Replaced the API with a single function:
```
sbp2rtcm_set_rtcm_out_mode(msm_type, &out_state);
```
With `MSM_UNKNOWN` interpreted as legacy obs setting. The buildroot PR making use of this interface is at https://github.com/swift-nav/piksi_buildroot/pull/926

nsirola:
There was a largish bug where the buffer length for the RTCM output callback was `u8` and thus any frame larger than 256 bytes caused strange results.

nsirola:
I've been running this build today and saw no issues. Will still run an overnight test to make sure, but barring any new issues I suppose these are good to go? @anth-cole

anth-cole:
Yep, looks good

PR #148, opened 2018-10-23
Update GNSS converters for Orion BDS implementation.

pgrgich:
One of a few submodules that has to be updated for https://swift-nav.atlassian.net/browse/MAP-684.


scarcanague:
Im confused, you are going back to a former Librtcm commit?

c/cmake/FindSbp.cmake
```diff
@@ -30,7 +30,7 @@ add_library(sbp SHARED IMPORTED GLOBAL)
 # This tells where the static libsbp binary will end up.  I have no
 # idea how to control this and just found it with `locate`.
 set_property(TARGET sbp
-        PROPERTY IMPORTED_LOCATION "${binary_dir}/src/libsbp-static.a")
+        PROPERTY IMPORTED_LOCATION "${binary_dir}/src/libsbp.a")
```
martin-swift: Why do we need this file at all if we are including sbp as a submodule?

pgrgich: I have no idea, I just did what made it build. :) @scarcanague could you comment?

pgrgich:
I'm pretty confused about which submodules are needed where. Trying to get it to run and once I know what I'm doing I'll push up the changes...

c/cmake/FindSbp.cmake
martin-swift: Ok - I removed this and its swift counterpart.  After updating the submodule pointers, it compiles fine :) I think we should remove those two cmake files.

c/include/gnss-converters/rtcm3_sbp.h
```diff
@@ -99,6 +99,10 @@ void rtcm2sbp_decode_frame(const uint8_t *frame,
                            uint32_t frame_length,
                            struct rtcm3_sbp_state *state);
 
+void rtcm2sbp_decode_payload(const uint8_t *payload,
+                             uint32_t payload_length,
+                             struct rtcm3_sbp_state *state);
```
pgrgich: This is called in /orion/src/gnss_converter.cc - it was deleted in https://github.com/swift-nav/gnss-converters/commit/0d2df057a3e1161570a5e6b2eb337c945eea8f55#diff-0320ba044265d3521efca007bbb4ce14, but I don't see how it can be avoided in Orion. Thoughts @nsirola @scarcanague ?

nsirola: Oops! Can you not use rtcm3sbp_decode_frame in Orion though? It looks like the whole frame is at hand here https://github.com/swift-nav/orion/blob/master/src/gnss_converter.cc#L61

pgrgich: That's the line that needs to remain `rtcm2sbp_decode_payload`, since `rtcm_frame.payload().data()` is obviously the payload and doesn't include the frame. Going back a few bytes in order to get the frame doesn't seem to be possible without changing `RtcmFrame`. @benjamin0 could you give your opinion?

benjamin0: I don't have all the context, but I think we need to keep `rtcm2sbp_decode_payload()` since we store the individual components of the RTCM frame - length, type, payload.

pgrgich: Yeah, that's the impression I was under from looking at the code. I'll put the function back into the interface, unless @nsirola objects.

c/cmake/FindSbp.cmake
martin-swift: I thought I pushed the removal of the deprecated cmake?  did you revert it?  should I make it a separate pull?

pgrgich: I haven't rebased yet. I will. :)

martin-swift: except I thought I pushed this up here already... lemme see what happened...

martin-swift: did you force-push?  either I failed to push, or you clobbered the commit.  Anyway, it's easy enough - just delete the FindSbp.cmake and FindSwiftNav.cmake files :)

c/include/gnss-converters/rtcm3_sbp.h
mfine: Right, I explicitly put that function in to break out the frame and payload. Thanks!

PR #149, opened 2018-10-24
Additional options for RTCM->SBP converter [STAR-531]

mbavaro:
Didn't make the Veoneer demo, because we use two different functions to create obs (one for legacy and one for MSM). Now fixed, could be a nice little tool for this kind of work.
Passing an option like `-b code:bias` adds pseudorange `bias` to the obs of code type `code`. 
Passing an option like `-c code:bias` adds FCN-dependent pseudorange `bias` to the obs of code type `code`. 
Passing an option like `-l code:bias` adds FCN-dependent carrier phase `bias` to the obs of code type `code`. 
Passing `-t` allows specifying the time (rather than picking it up from current PC time)


c/src/biases.c
```diff
@@ -0,0 +1,69 @@
...
+#include <assert.h>
+
+#include <rtcm3/constants.h>
+#include <rtcm3/msm_utils.h>
+#include <swiftnav/constants.h>
+#include <swiftnav/signal.h>
+
+#include "rtcm3_msm_utils.h"
+
+float sbp_signal_biases[CODE_COUNT] = {0.0f};
```
nsirola: Should these be `static` and have proper setters?

mbavaro: yeah, makes sense, let me do it..

mbavaro: actually... I thought about it and having a clean set/get  interface adds almost as much code as this PR does.. do you really feel strongly about variables exported by `biases.h`?

nsirola: Well not strongly opposed, just feels a bit icky when these biases will now be applied by default and it's not so easy to see what their scope is and where they come from. I take it they will still be zeros in every other case except when running the `rtcm32sbp` tool from command line and setting them explicitly.

silverjam:
@anth-cole do we still need this?

anth-cole:
@silverjam thanks for tagging this! I didn't realize this existed and I was about to start adding similar functionality. I'll make some changes here and then try and land it.

PR #152, opened 2018-10-25
Output empty age of corrections field for SBAS positions

anth-cole:
Currently, we're outputting 6553.3s for SBAS positions in the age of correction field, this sets it to an empty field.

### DIFF
--- a/c/src/nmea.c
+++ b/c/src/nmea.c
@@ -344,8 +344,9 @@ void send_gpgga(const struct sbp_nmea_state *state) {
     NMEA_SENTENCE_PRINTF(",,,M,,M,");
   }
 
-  if ((fix_type == NMEA_GGA_QI_DGPS) || (fix_type == NMEA_GGA_QI_FLOAT) ||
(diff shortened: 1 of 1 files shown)


c/src/nmea.c
```diff
@@ -344,8 +344,9 @@ void send_gpgga(const struct sbp_nmea_state *state) {
     NMEA_SENTENCE_PRINTF(",,,M,,M,");
   }
 
-  if ((fix_type == NMEA_GGA_QI_DGPS) || (fix_type == NMEA_GGA_QI_FLOAT) ||
-      (fix_type == NMEA_GGA_QI_RTK)) {
+  if ((fix_type == NMEA_GGA_QI_DGPS &&
+       ((sbp_pos_llh->flags & POSITION_MODE_MASK) != POSITION_MODE_SBAS)) ||
+      (fix_type == NMEA_GGA_QI_FLOAT) || (fix_type == NMEA_GGA_QI_RTK)) {
```
nsirola: I wonder if we could unit test this

PR #153, opened 2018-10-26
EX-524, EX-534: refactor NMEA sending logic, allow GSA to use latest obs message

nsirola:
Issue https://swift-nav.atlassian.net/browse/EX-534

Refactor the NMEA sending to a bit more extensible framework and eliminate repeated code.

Fixes also issue https://swift-nav.atlassian.net/browse/EX-524 by allowing the GSA message to be sent as long as there is an entire `MSG_OBS` sequence received with time stamp not older than the current solution epoch. This will enable GSA messages also in Time Matched mode where solution and DOP come with a latency, but not interfere with Low Latency operation.

Also found and fixed a major units blunder in `sbp_gpsdifftime`, luckily it was so far used only for checking if two time stamps are equal for which it did work correctly.

## Testing

The NMEA unit tests pass although they are somewhat cursory.

Verified with the buildroot PR https://github.com/swift-nav/piksi_buildroot/pull/1158 on bench in both Low Latency (1Hz and 5Hz) and Time Matched (1Hz) that all NMEA messages get sent at the requested rates.


c/src/sbp_nmea.c
```diff
@@ -10,125 +10,95 @@
...
-  }
-  return false;
+static bool gsa_ready(const struct sbp_nmea_state *state) {
+  /* Ready to send when the DOP timestamp matches with current epoch, and there
+   * is a full obs message sequence not older than the solution epoch.
+   * (This enables this message also in Time Matched mode, although the list of
+   * active satellites will be that of current obs and not the earlier ones that
+   * the solution and DOP are based on.) */
+  return state->sbp_utc_time.tow == state->sbp_dops.tow &&
+         state->sbp_utc_time.tow <= state->obs_time.tow &&
```
nsirola: This will miss a GSA message in the last seconds of the week if observations have already come in with their timestamp on the next week. Not entirely trivial to fix as the utc time message does not carry the GPS week number.

nsirola: Fixed

PR #154, opened 2018-11-01
EX-415: Shorten MSM timeout to 5 s

nsirola:
Currently when switching from MSM RTCM input stream to legacy stream, there's a 60 second timeout during which observations are silently dropped, which might be confusing for the unsuspecting user.

As the purpose of this timeout was just to suppress the legacy observations in a stream that mixes both MSM and legacy, a much shorter delay should be sufficient.


nsirola:
Tested with the buildroot test PR https://github.com/swift-nav/piksi_buildroot/pull/985 with a Piksi-to-Piksi pair that switching between RTCM OUT types is handled gracefully in the receiver. Switch from Legacy to MSM is seamless and stays in Fixed mode, but when switching from MSM to Legacy, the solution fir coasts in Fixed during the 5 second timeout but then drops to Float for a couple of seconds when base observations resume. (probably not an issue)

PR #156, opened 2018-11-08
Add populating FCN to RTCM1020

mbavaro:
To be tested.


c/src/rtcm3_sbp.c
```diff
@@ -1505,9 +1506,10 @@ void rtcm3_msm_to_sbp(const rtcm_msm_message *msg,
       if (msg->header.cell_mask[sat * num_sigs + sig]) {
         sbp_gnss_signal_t sid;
         const rtcm_msm_signal_data *data = &msg->signals[cell_index];
-        if (get_sid_from_msm(&msg->header, sat, sig, &sid, state) &&
-            data->flags.valid_pr && data->flags.valid_cp &&
-            !unsupported_signal(&sid)) {
+        bool got = get_sid_from_msm(&msg->header, sat, sig, &sid, state);
```
nsirola: Maybe `sid_valid` would be more illustrative name

c/src/rtcm3_sbp.c
```diff
@@ -1815,7 +1817,7 @@ static void rtcm_init_obs_message(rtcm_obs_message *msg,
     msg->header.msg_num = 1012;
     /* GLO epoch time DF034 uint32 27 */
     msg->header.tow_ms =
-        (u32)(compute_glo_tod(state->sbp_header.t.tow, state) * S_TO_MS);
+        (u32) rint(compute_glo_tod(state->sbp_header.t.tow, state) * S_TO_MS);
```
nsirola: Good catch!

c/src/rtcm3_sbp.c
```diff
@@ -227,6 +227,7 @@ void rtcm2sbp_decode_payload(const uint8_t *payload,
       if (RC_OK == rtcm3_decode_glo_eph(&payload[byte], &msg_eph)) {
         msg_ephemeris_glo_t sbp_glo_eph;
         rtcm3_glo_eph_to_sbp(&msg_eph, &sbp_glo_eph, state);
+        rtcm2sbp_set_glo_fcn(sbp_glo_eph.common.sid, sbp_glo_eph.fcn, state);
```
nsirola: The only thing I'd check (if I had a specification including the 1020 message at hand now) was if it is possible for the FCN value to be unknown/undefined, and that it gets encoded correctly into `SBP_GLO_FCN_UNKNOWN` before this point.

nsirola: This version FCN seems to go up to eleven
image

nsirola: Ok `rtcm2sbp_set_glo_fcn` uses `sbp_fcn_to_rtcm` and that will properly store too large values as unknown FCNs, so this should work allright.

PR #159, opened 2018-11-16
EX-445: Send 1006, 1008, 1033, add API to set ant&recv decriptors

nsirola:
Issue: https://swift-nav.atlassian.net/browse/EX-445

- Send 1006 base pos message instead of 1005, using `antenna_height` as set by buildroot (defaults to 0.0).
- Send 1008 and 1033 recv&antenna descriptor messages, with values set by buildroot (defaults to blanks).
- Use `MEMCPY_S` instead of `memcpy` so all bounds are checked.
- Fix discrepancies in RTCM station ID

Pull in https://github.com/swift-nav/librtcm/pull/67

Revised design doc at https://swift-nav.atlassian.net/wiki/spaces/EXF/pages/290062688/RTCM+Out

Used by buildroot PR https://github.com/swift-nav/piksi_buildroot/pull/1010


c/src/rtcm3_sbp.c
```diff
@@ -874,9 +884,65 @@ void rtcm3_1006_to_sbp(const rtcm_msg_1006 *rtcm_1006,
...
+
+  /* Reference Station ID DF003 uint12 12 */
+  rtcm_1008->msg_1007.stn_id = rtcm_1033->stn_id;
+  /* Descriptor Counter N DF029 uint8 8 N <= 31 */
+  rtcm_1008->msg_1007.ant_descriptor_counter =
+      rtcm_1033->ant_descriptor_counter;
+  /* Antenna Descriptor DF030 char8(N) 8*N */
+  memcpy(rtcm_1008->msg_1007.ant_descriptor,
+         rtcm_1033->ant_descriptor,
+         rtcm_1033->ant_descriptor_counter);
```
pmiettinen: Is the length validated somewhere?

nsirola: Adding asserts here as well.
This ultimately comes from buildroot `sbp_rtcm3_bridge` that should use the same max length for these strings.

nsirola: Ended up replacing all `memcpy`s with `MEMCPY_S`

c/src/rtcm3_sbp.c
```diff
@@ -103,6 +104,8 @@ void sbp2rtcm_init(struct rtcm3_out_state *state,
   state->send_msm_obs = true;
   state->msm_type = MSM5;
 
+  state->ant_height = 0.0;
```
anth-cole: Do we also need to init the antenna and receiver strings?

c/src/rtcm3_sbp.c
```diff
@@ -1600,11 +1692,25 @@ void sbp2rtcm_base_pos_ecef_cb(const u16 sender_id,
...
-  u16 frame_size = encode_rtcm3_frame(&msg_1005, 1005, frame);
+
+  /* generate and send the base position message */
+  sbp_to_rtcm3_1006(
+      (const msg_base_pos_ecef_t *)msg, sender_id, &msg_1006, state);
+  u16 frame_size = encode_rtcm3_frame(&msg_1006, 1006, frame);
+  state->cb_sbp_to_rtcm(frame, frame_size, state->context);
+
+  /* generate and send the receiver and antenna description messages */
+  generate_rtcm3_1033(sender_id, &msg_1033, state);
```
anth-cole: any reason to pass sender_id as a separate argument? isn't it a field in the state struct already?

c/src/rtcm3_sbp.c
nsirola: Would not hurt, added memsets for those.

c/src/rtcm3_sbp.c
nsirola: Good point. And checking all the uses of sender id, I noticed some discrepancies in conversions between SBP and RTCM IDs, so fixed those and clarified the notation so that SBP has **sender** ID and RTCM **station** ID.

PR #161, opened 2018-12-05
EX-472: Verify RTCM antenna height setting

nsirola:
Pull in https://github.com/swift-nav/librtcm/pull/70
Modify the antenna height setter to check for valid value, and to return boolean result.


c/include/gnss-converters/rtcm3_sbp.h
```diff
@@ -139,7 +139,7 @@ void sbp2rtcm_set_glo_fcn(sbp_gnss_signal_t sid,
                           u8 sbp_fcn,
                           struct rtcm3_out_state *state);
 
-void sbp2rtcm_set_ant_height(double ant_height, struct rtcm3_out_state *state);
+bool sbp2rtcm_set_ant_height(double ant_height, struct rtcm3_out_state *state);
```
pmiettinen: Is the caller checking the return value?

nsirola: Only one caller at the moment (sbp_rtcm3_bridge in buildroot). Modifying that will be the next PR.

PR #163, opened 2018-12-11
Standalone SBP converter

mfine:
Provide a standalone SBP converter that contains all of the state necessary to converter SBP frames to RTCM bytes.


c/include/gnss-converters/sbp_converter.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+#ifndef GNSS_CONVERTERS_SBP_CONVERTER_INTERFACE_H
+#define GNSS_CONVERTERS_SBP_CONVERTER_INTERFACE_H
+
+#include <gnss-converters/rtcm3_sbp.h>
+#include <swiftnav/fifo_byte.h>
+#include <libsbp/sbp.h>
+
+#define BUFFER_SIZE 8192
```
mbavaro: let's add a comment that this HAS to be a power of two.

c/include/gnss-converters/sbp_converter.h
```diff
@@ -0,0 +1,46 @@
...
+
+struct sbp_converter {
+  struct rtcm3_out_state state;
+  fifo_t fifo;
+  sbp_state_t sbp_state;
+  sbp_msg_callbacks_node_t base_pos_ecef_callback_node;
```
mbavaro: not entirely up to speed on this one.. why does the base need to be a state of the converter?

mfine:
Using this in a very limited manner as a set of self-contained conversions. Will likely need some enhancements going forward. Bringing in for now.

PR #164, opened 2018-12-13
Antenna known flag to restrict 1033 and 1008 messages

mfine:
Only send out 1033 and 1008 messages if the antenna is known.

For VRS generated observations, we are currently translating just observations and positions and not providing any antenna or receiver descriptions. The presence of empty antenna and receiver descriptions in 1033 and 1008 over the network is not useful. Restrict their sending based on whether actual antenna data has been set.


mfine:
@anth-cole sounds good - I'd like to even in that case, not send these messages by default, and only when the software has explicitly set the receiver to the NULL antenna.

PR #166, opened 2019-01-14
EX-499, EX-501: Retain millisecond part when converting GLO TOW

nsirola:
GLO time stamps are currently rounded down to a whole second, which breaks RTCM to SBP conversion when observation rate is larger than 1Hz. (see https://swift-nav.atlassian.net/browse/EX-499)

Fixed by handling the TOW as milliseconds instead of seconds. `s32` has plenty of range (+-3.5 weeks in milliseconds)

Update: the fix uncovered another bug in MSM encoder where GLO TOW was truncated down to millisecond instead of rounded correctly (sometimes resulting TOW x.399 instead of .400) (EX-501)

## Testing

- [x] Unit test with a snippet of RTCM stream recorded from Piksi at 2Hz. Fails with master but passes with this PR.
- [x] Bench test (with the build in https://github.com/swift-nav/piksi_buildroot/pull/1120): Both Legacy and MSM work with all solution frequencies


nsirola:
Tested both in unit tests and bench, merging.

PR #167, opened 2019-01-21
Fix rounding issue

mbavaro:
Incredible how this can have slipped under the radar?
https://swift-nav.atlassian.net/browse/STAR-396

### DIFF
--- a/c/src/rtcm3_sbp.c
+++ b/c/src/rtcm3_sbp.c
@@ -1519,7 +1519,7 @@ void add_msm_obs_to_buffer(const rtcm_msm_message *new_rtcm_obs,
 
     /* Build an SBP time stamp */
     new_sbp_obs->header.t.wn = obs_time.wn;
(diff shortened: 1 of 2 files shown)


anth-cole:
Lgtm - seems amazing we haven't seen this before

anth-cole:
Unless it indicates we're usually just ahead of the for epoch time?

mbavaro:
Turns out, 131789800 ms is a very power-y of two number:
```
num2hex(131789800)
ans = 419f6bcfa0000000
```
and so the division by 1000 corrupts it "heavily"
```
num2hex(131789800 / 1000)
ans = 4100166e66666666
```
to the point that
```
printf("%.16f\n", hex2num(num2hex(131789800 / 1000)))
131789.7999999999883585
```

Go figure, really, as we haven't exercised this decoder enough? 😄

nsirola:
Opened a new issue EX-515 for fixing the remaining problems of this type

PR #169, opened 2019-01-23
Set the antenna for the conversion

mfine:
Set then antenna on conversion. This should be more general, but right now it's just being used by the caster service and I don't want to thread these string through.


mfine:
Should we use [ADVNULLANTENNA](https://www.use-snip.com/kb/knowledge-base/what-is-advnullantenna/) instead?

benjamin0:
@mfine Not sure, `ADVNULLANTENNA` does look more appropriate - maybe we could do a quick static test with the Javad with both?

mfine:
@benjamin0 @jaredw42 validated with it as is. Happy to leave it. I don't find `ADVNULLANTENNA` in our `igs14.atx`

PR #168, opened 2019-01-22
EX-515: fix all compiler warnings on float conversion

nsirola:
Should take care of the rest of hidden rounding problems...

Fixes https://swift-nav.atlassian.net/browse/STAR-395 (the RTCM snippet from the issue added as an unit test)

PR #172, opened 2019-02-08
EX-516: enable -Wfloat-conversion warning, update libswiftnav

nsirola:
Follow up on https://github.com/swift-nav/gnss-converters/pull/168 and enable the compiler warning to catch these in the future.

Update `librtcm` to the branch with warnings fixed.


nsirola:
The docker build or something does not seem to like the `libswiftnav` branch update?

PR #173, opened 2019-02-14
EX-534: Rework sbp2nmea

pmiettinen:
I wanted to try some stuff and ended up with this rework. Maybe there's something useful. Not tested other than unit tests pass.

### DIFF
--- a/c/src/sbp_nmea.c
+++ b/c/src/sbp_nmea.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Swift Navigation Inc.
+ * Copyright (C) 2018-2019 Swift Navigation Inc.
  * Contact: Swift Navigation <dev@swiftnav.com>
  *
  * This source is subject to the license found in the file 'LICENSE' which must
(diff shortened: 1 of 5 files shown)


c/src/sbp_nmea.c
```diff
@@ -207,72 +226,40 @@ void sbp2nmea_obs(const msg_obs_t *sbp_obs,
...
+void sbp2nmea_soln_freq_set(sbp2nmea_t *state, float soln_freq) {
+  state->soln_freq = soln_freq;
 }
 
-void sbp2nmea_set_gpgll_rate(const int gpgll_rate,
-                             struct sbp_nmea_state *state) {
-  state->gpgll_rate = gpgll_rate;
-}
+sbp2nmea_t *sbp2nmea_init(void (*cb_sbp_to_nmea)(u8 msg_id[])) {
+  sbp2nmea_t *state = calloc(1, sizeof(sbp2nmea_t));
```
nsirola: This is the only thing that worries me a bit

nsirola:
Yes overall this is greatly improves the structure and removes much copy-pasted code. It's just the run-time memory allocation go against the pattern in the other SBP-to-X bridges... Though the caller already has to explicitly both initialize and clean up the converters, so I guess this wouldn't be that much of a leap.

c/src/sbp_nmea.c
pmiettinen: I have few different ideas to make this stack based. They are trade-offs on the hackerino-memoryfootprint axis though. I'll see if a pleasing compromise can be found.

pmiettinen:
Dynamic memory allocation is now gone. Downsides:
- data structures are exposed in the header for the compiler know the sizes
- memory footprint is larger than necessary as `sbp_state_entry_t` size based on the largest SBP message; could be worked around by allocating just a plumb of `u8` array size sum of message sizes but working with it gets too hacky IMO

nsirola:
Ok this works in the buildroot test build, will pull this into my branch and continue from there. Thanks!

PR #174, opened 2019-02-18
Update libswiftnav, update C build to gcc-6

nsirola:
Update `libswiftnav` to the tip of master. 

This required some changes to the environment:
- restrict submoduling to just one level as Travis doesn't seem to have permission to `libswiftnav/docker-recipes`
- update the C build to `gcc-6` (following most other repos)
- set up separate matrix build for Haskell to keep it in `gcc-4.8` (it doesn't like `gcc-6`)
- (also the manual `cmake` install is not needed anymore)

This branch also compiles in `piksi-buildroot`: https://github.com/swift-nav/piksi_buildroot/pull/1159


mfine:
What was the haskell issue? There will be more issues if it gets updated.

mfine:
Yuck, https://github.com/commercialhaskell/stack/issues/3472

nsirola:
> Yuck, [commercialhaskell/stack#3472](https://github.com/commercialhaskell/stack/issues/3472)

Yep, that one. It happens also with `lts-11.21`. With `lts-13.7` there was different error `No instance for (unliftio-core-0.1.2.0:Control.Monad.IO.Unlift.MonadUnliftIO(ConvertT Store IO)) arising from a use of ‘runConduitRes’`

I'll try if I can get Haskell to link with the default `gcc-4.8` and use `gcc-6` only for the C part...

nsirola:
Ok managed to set up the matrix build so the builds go through

PR #179, opened 2019-03-25
Uses system define install paths

woodfell:
Use the cmake module GNUInstallDirs to define standard system install paths. This module works well when providing an installation prefix or destdir to cmake/make. This will replace all hard coded installation paths


jbangelo:
One thing to note, installing this also installs libsbp and libswiftnav. Is that intentional?

PR #188, opened 2019-04-19
Updated CMake to not include dependencies if they have already been defined elsewhere

jbangelo:
Changed CMake to only call `add_subdirectory()` if a dependency hasn't already been defined.

Switched to using `PROJECT_SOURCE_DIR` when adding a subdirectory, which won't break when this repo is submoduled.


c/CMakeLists.txt
```diff
@@ -9,16 +9,16 @@ add_compile_options(-UNDEBUG)
...
 endif()
 
-if(EXISTS ${CMAKE_SOURCE_DIR}/libsbp/c)
+if(NOT TARGET sbp AND EXISTS ${PROJECT_SOURCE_DIR}/libsbp/c)
   include_directories("${PROJECT_SOURCE_DIR}/libsbp/c/include")
   add_subdirectory(libsbp/c)
 endif()
 
-if(EXISTS ${CMAKE_SOURCE_DIR}/libswiftnav)
+if(NOT TARGET swiftnav AND EXISTS ${PROJECT_SOURCE_DIR}/libswiftnav)
```
martin-swift: I believe we actually want to test for the existence of the CMakeLists.txt, no?  The directory will get created upon a clone, but the CMakeLists.txt will only be created after a `git submodule update --init --recursive`  @woodfell ?

c/CMakeLists.txt
```diff
@@ -9,16 +9,16 @@ add_compile_options(-UNDEBUG)
...
 set(CMAKE_C_FLAGS "-Wall -Wextra -Wno-strict-prototypes -Werror -std=gnu99 -fno-unwind-tables -fno-asynchronous-unwind-tables -Wimplicit -Wshadow -Wswitch-default -Wswitch-enum -Wundef -Wuninitialized -Wcast-align -Wformat=2 -Wimplicit-function-declaration -Wredundant-decls -Wformat-security -Wfloat-conversion -ggdb ${CMAKE_C_FLAGS}")
 
-if(EXISTS ${CMAKE_SOURCE_DIR}/librtcm/c)
+if(NOT TARGET rtcm AND EXISTS ${PROJECT_SOURCE_DIR}/librtcm/c)
   add_subdirectory(librtcm/c)
 endif()
 
-if(EXISTS ${CMAKE_SOURCE_DIR}/libsbp/c)
+if(NOT TARGET sbp AND EXISTS ${PROJECT_SOURCE_DIR}/libsbp/c)
   include_directories("${PROJECT_SOURCE_DIR}/libsbp/c/include")
```
martin-swift: would be great if we could also get rid of this - is libsbp exporting its includes properly yet?

jbangelo: Good question! Looks like it.

c/CMakeLists.txt
woodfell: @martin-swift Correct, this is a long standing bug in this makefile. We need to always check for the presence of CMakeLists.txt rather than the directory for exactly the reason you specified. 

This exact bug is also present when including libsbp as well, but it is hidden because the code in libsbp is actually in a subdirectory within the submodule. So it the submodule isn't updated then the subdirectory which is under test will not exist and the conditional will behave as expected.

PR #189, opened 2019-04-20
Harmonize WN and TOW in Ephemeris handler [ORI-482]

mbavaro:
I don't know if this is the only place.. but like this it works better in my head


c/src/rtcm3_sbp_ephemeris.c
```diff
@@ -258,8 +256,8 @@ void rtcm3_gps_eph_to_sbp(rtcm_msg_eph *msg_eph,
   sbp_gps_eph->iode = msg_eph->kepler.iode;
   sbp_gps_eph->iodc = msg_eph->kepler.iodc;
 
-  sbp_gps_eph->toc.wn = (state->time_from_rover_obs.wn & 0xFC00) + msg_eph->wn;
```
mbavaro: the time from the rover obs does not have a week of its own in RTCM.. and there is a thought of actually deriving it from the ephemerides, so this loop makes me uncomfortable...

c/src/rtcm3_sbp_ephemeris.c
```diff
@@ -258,8 +256,8 @@ void rtcm3_gps_eph_to_sbp(rtcm_msg_eph *msg_eph,
...
-  sbp_gps_eph->toc.tow = (u32)rint(msg_eph->kepler.toc * GPS_TOC_RESOLUTION);
+  sbp_gps_eph->toc.wn = gps_adjust_week_cycle(msg_eph->wn, GPS_WEEK_REFERENCE);
+  sbp_gps_eph->toc.tow = msg_eph->kepler.toc * GPS_TOC_RESOLUTION;
```
mbavaro: I think it's all integers?

c/src/rtcm3_sbp_ephemeris.c
```diff
@@ -303,15 +301,9 @@ void rtcm3_glo_eph_to_sbp(rtcm_msg_eph *msg_eph,
...
-   * bits */
-  sbp_gal_eph->common.toe.tow =
-      (u32)rint(msg_eph->toe * GALILEO_TOC_RESOLUTION);
-  sbp_gal_eph->common.toe.wn =
-      gps_adjust_week_cycle(msg_eph->wn, GPS_WEEK_REFERENCE);
-  /* gps_time_match_weeks( TODO(ORI-482): as is, this updates a temporary - removing for now. */
-  /*     &(gps_time_t){sbp_gal_eph->common.toe.tow, sbp_gal_eph->common.toe.wn}, */
-  /*     &state->time_from_rover_obs); */
+  /* Galileo week is 12 bit (DF289) */
+  sbp_gal_eph->common.toe.wn = gps_adjust_week_cycle(msg_eph->wn, GPS_WEEK_REFERENCE);
```
mbavaro: the Galileo week will rollover every 80 years, but we can still say that if it's smaller than our reference we should add 1024 (~20 years)

c/src/rtcm3_sbp_ephemeris.c
```diff
@@ -347,26 +339,20 @@ void rtcm3_gal_eph_to_sbp(rtcm_msg_eph *msg_eph,
...
-  u32 tow_ms = (u32)rint(msg_eph->toe * BEIDOU_TOC_RESOLUTION * SECS_MS);
-  beidou_tow_to_gps_tow(&tow_ms);
   gps_time_t toe;
-  compute_gps_message_time(tow_ms, &toe, &state->time_from_rover_obs);
-  sbp_bds_eph->common.toe.wn = toe.wn;
-  sbp_bds_eph->common.toe.tow = (u32)rint(toe.tow);
-
+  /* Beidou week is 13 bit (DF489) */
+  toe.wn = msg_eph->wn;
+  toe.tow = (msg_eph->toe * BEIDOU_TOE_RESOLUTION) + BDS_SECOND_TO_GPS_SECOND;
```
mbavaro: Beidou is a bit special because it has 14 seconds offset with GPS.. as its `t0` is when GPS had 14 UTC leap seconds.. so the idea here is:
1. add them, 
2. normalize, 
3. adjust week

c/src/rtcm3_sbp_ephemeris.c
```diff
@@ -258,8 +258,8 @@ void rtcm3_gps_eph_to_sbp(rtcm_msg_eph *msg_eph,
   sbp_gps_eph->iode = msg_eph->kepler.iode;
   sbp_gps_eph->iodc = msg_eph->kepler.iodc;
 
-  sbp_gps_eph->toc.wn = (state->time_from_rover_obs.wn & 0xFC00) + msg_eph->wn;
-  sbp_gps_eph->toc.tow = (u32)rint(msg_eph->kepler.toc * GPS_TOC_RESOLUTION);
+  sbp_gps_eph->toc.wn = gps_adjust_week_cycle(msg_eph->wn, GPS_WEEK_REFERENCE);
+  sbp_gps_eph->toc.tow = msg_eph->kepler.toc * GPS_TOC_RESOLUTION;
 }
 
 void rtcm3_glo_eph_to_sbp(rtcm_msg_eph *msg_eph,
```
mbavaro: actually, this one is flawed..
using `compute_glo_time()` below will lead to setting the `wn` to `0xFFFF` (invalid) pretty much always
(see https://swift-nav.atlassian.net/projects/STAR/issues/STAR-572)
as the boundary for time validity is `GLO_SANITY_THRESHOLD_S`, that is 30 seconds.
I think we should use RTCM eph only parameters, I'll see if it's possible to resort to the `NT` field and use some LSN function to get `wn` reliably here.

nsirola: Made a small PR for this against this branch in https://github.com/swift-nav/gnss-converters/pull/190

PR #193, opened 2019-04-29
Import standard cmake modules, standardise dependency resolution [ESD-1246] [ESD-1247]

woodfell:
Integrates with swift-nav/cmake#2

Import common cmake repository. Set up package cmake options to enable/disable unit tests.

This project depends on the following repositories and related PRs

libsbp https://github.com/swift-nav/libsbp/pull/692
libswiftnav - https://github.com/swift-nav/libswiftnav/pull/85
librtcm https://github.com/swift-nav/librtcm/pull/85


benjaminaltieri:
I think travis needs to do the submodule pull in order for the new options/methods to work. For some reason it was working without before..

PR #194, opened 2019-05-01
Moved the core logic of the rtcm3tosbp program into a function in the gnss-converters library

jbangelo:
The intention of this PR is to make the core RTCM->SBP loop available in the `gnss-converters` library. This loop is useful in other projects that want to internally perform the RTCM->SBP conversion without having to spin up a new process and route data through it. The loop also contains a fair amount of RTCM specific knowledge which is best maintained with the rest of the RTCM->SBP code, and not in external projects.

Unanswered questions:
1. The reading function pointer isn't in the `rtcm3_sbp_state` since it's not a core need for the conversion but it uses the `context` pointer from the state struct. Does this make sense?
2. The internal fifo and buffer sizes are hard coded, should they be made configurable by the caller?


c/src/rtcm3_sbp.c
```diff
@@ -2280,3 +2281,92 @@ void sbp2rtcm_sbp_osr_cb(const u16 sender_id,
...
+  uint8_t fifo_buf[FIFO_SIZE] = {0};
+  fifo_t fifo;
+  fifo_init(&fifo, fifo_buf, sizeof(fifo_buf));
+
+  uint8_t inbuf[BUFFER_SIZE];
+  ssize_t numread;
+  while ((numread = read_func(inbuf, BUFFER_SIZE, state->context)) > 0) {
+    ssize_t numwritten = fifo_write(&fifo, inbuf, numread);
+    if (numwritten != numread) {
+      fprintf(stderr,
```
martin-swift: if this is intended to be called by things that are not stand-alone programs, do we really want this trying to write to stderr?  maybe we should change that to something that binds it to one of our log_ functions for embedded, and stderr if it's a command-line program...

jbangelo: I agree, but the rest of the library prints straight to stderr. I'd prefer to leave switching to our logging functions for a separate PR.

c/src/rtcm3_sbp.c
```diff
@@ -2280,3 +2281,92 @@ void sbp2rtcm_sbp_osr_cb(const u16 sender_id,
   u16 frame_size = encode_rtcm3_frame(&osr_msg, 4062, frame);
   state->cb_sbp_to_rtcm(frame, frame_size, state->context);
 }
+
+static uint16_t extract_msg_len(uint8_t *buf) {
```
nsirola: could use `const uint8_t *bug`

c/src/rtcm3_sbp.c
```diff
@@ -2280,3 +2281,92 @@ void sbp2rtcm_sbp_osr_cb(const u16 sender_id,
...
+  return (((uint16_t)buf[1] << 8) | buf[2]) & RTCM3_MAX_MSG_LEN;
+}
+
+/* buf_len is your total allocated space - can be much bigger than
+   your actual message */
```
nsirola: consider rewriting the comment e.g.  `your` -> `the`

c/src/rtcm3_sbp.c
```diff
@@ -2280,3 +2281,92 @@ void sbp2rtcm_sbp_osr_cb(const u16 sender_id,
...
+            computed_crc);
+  }
+  return (frame_crc == computed_crc);
+}
+
+/* You may reduce FIFO_SIZE if you need a lower memory footprint. */
+#define FIFO_SIZE (1 << 14)
+#define BUFFER_SIZE (FIFO_SIZE - RTCM3_MSG_OVERHEAD - RTCM3_MAX_MSG_LEN)
+
+void rtcm2sbp_process(struct rtcm3_sbp_state *state, int (*read_func)(uint8_t *buf, size_t len, void *ctx)) {
```
nsirola: Now that this is removed from the context of the standalone converter, it may be worth adding a short comment to this function describing the inputs, and mentioning that the output happens through the callback in `state`

jbangelo:
@anth-cole Good question. Those are definitely options, an alternative is to use variable length arrays which C supports and uses the stack. VLAs are not a required feature in C11, gcc and clang have support but most notably MSVC does not. VLAs also mean more machine code is generated, the main complaint against them.

I would suggest making it a compile time configuration value.

martin-swift:
Note that the FIFO_SIZE has to be a power of 2.  The fifo implementation uses some clever bit roll-over tricks for correct operation.

jbangelo:
Merging this as is. If we feel the need to make the buffer sizes configurable we can make a separate PR. The existing buffer sizes is fine for my purposes.

PR #200, opened 2019-05-10
Added functionality to convert the separate SSR orbit and clock RTCM messages into the combined SRR orbit and clock SBP message

jbangelo:


c/include/gnss-converters/rtcm3_sbp.h
```diff
@@ -54,6 +54,18 @@ typedef enum {
...
 
+// A struct for storing either an SSR orbit correction message
+// or an SSR clock correction message. Used for combining the
+// separate messages into a combined message
+typedef struct {
+  union {
+    rtcm_msg_clock clock;
+    rtcm_msg_orbit orbit;
+  };
+  bool populated;
```
benjamin0: What does `populated` mean here? Should we have a `contains_clock` and `contains_orbit`?

c/src/rtcm3_sbp.c
```diff
@@ -320,6 +320,67 @@ void rtcm2sbp_decode_payload(const uint8_t *payload,
...
     }
+
+    /* The following two chunks of messages handle converting separate SSR orbit correction
+     * and SSR clock correction messages into a single SBP message containing both orbit and
+     * clock corrections.
+     * This code makes the following assumptions:
+     *  1. Each pair of orbit and clock messages contain data for the same group of satellites,
+     *     this implies that we don't support the multimessage flag in the RTCM message. We do
+     *     match the clock correction satid with the orbit correction satid, so the corrections
+     *     don't have to be in the same order but we will silently drop any correction that we
```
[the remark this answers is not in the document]
jbangelo: We certainly can, though it's not clear to me how or if the rest of gnss-converters logs messages on errors. It looks like in a few cases it prints to `stderr`, but that doesn't make a ton of sense if it's being used as a library.

c/include/gnss-converters/rtcm3_sbp.h
jbangelo: `populated` means that there is valid data stored, and `contains_clock` determines if the data is either clock or orbit data. My thinking was that this approach doesn't have invalid states, where using `contains_clock` and `contains_orbit` has the case of if they are both true which one to believe?

I don't have strong feelings either way, if you think the `contains_clock` and `contains_orbit` approach is easier to read we can go with it.

scarcanague: Yeah I would personally be in favor of `contains_orbit` it's seems more natural to me

c/src/rtcm3_sbp_ssr.c
```diff
@@ -40,78 +41,134 @@ gps_time_sec_t compute_ssr_message_time(
...
+  uint8_t length;
+  msg_ssr_orbit_clock_t *sbp_orbit_clock = (msg_ssr_orbit_clock_t *)buffer;
 
-    sbp_orbit_clock->dot_radial = msg_orbit_clock->orbit[sat_count].dot_radial;
-    length += sizeof(sbp_orbit_clock->dot_radial);
+  for (int sat_count = 0; (sat_count < msg_clock->header.num_sats) && (sat_count < msg_orbit->header.num_sats); ++sat_count) {
+    memset(buffer, 0, sizeof(buffer));
+    length = 0;
+
+    // We aren't guarnteed that the satellite info in the orbit and clock messages is in the same order
```
scarcanague: Nit: comment has a few typos

jbangelo:
Talked with @benjamin0 , and this isn't required any more. @scarcanague when you have some time we should chat about if we want this merged in anyways.

PR #201, opened 2019-05-14
Update LSN pointer

mbavaro:


c/CMakeLists.txt
```diff
@@ -7,7 +7,11 @@ include(GNUInstallDirs)
 add_compile_options(-UNDEBUG)
 
 # Some compiler options used globally
-set(CMAKE_C_FLAGS "-Wall -Wextra -Wno-strict-prototypes -Werror -std=gnu99 -fno-unwind-tables -fno-asynchronous-unwind-tables -Wimplicit -Wshadow -Wswitch-default -Wswitch-enum -Wundef -Wuninitialized -Wcast-align -Wformat=2 -Wimplicit-function-declaration -Wredundant-decls -Wformat-security -Wfloat-conversion -ggdb ${CMAKE_C_FLAGS}")
+set(CMAKE_C_FLAGS "-Wall -Wextra -Wno-strict-prototypes -Werror -std=gnu99 -fno-unwind-tables -fno-asynchronous-unwind-tables -Wimplicit -Wshadow -Wswitch-default -Wswitch-enum -Wundef -Wuninitialized -Wcast-align -Wformat=2 -Wimplicit-function-declaration -Wredundant-decls -Wformat-security -ggdb ${CMAKE_C_FLAGS}")
+
+if (CMAKE_C_COMPILER_VERSION VERSION_GREATER 5.0)
```
mbavaro: temporary fix as building with gcc 4.7 (official raspberry-pi cross toolchain) is broken otherwise

jbangelo: Is the problem that gcc 4.7 has incorrectly flags float conversions?

mbavaro: doesn't seem to know `-Wfloat-conversion` at all..

jbangelo:
```
if (NOT "${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU" OR CMAKE_C_COMPILER_VERSION VERSION_GREATER 5.0)
```
This would limit the exclusion to just gcc below version 5.0, we'd still want `clang` to have this warning enabled.

mbavaro: oh dear, so we need to change here too:
https://github.com/swift-nav/libswiftnav/blob/master/CMakeLists.txt#L93
and
https://github.com/swift-nav/librtcm/blob/master/c/src/CMakeLists.txt#L49

PR #203, opened 2019-05-15
When the determinant and the square of the trace are very close, taki…

anth-cole:
…ng the difference

using double precision can sometimes lead to very small negative numbers. Taking the
fabs will ensure this number is positive, for these tiny numbers the overall effect
is negligable


c/src/nmea.c
```diff
@@ -940,15 +940,15 @@ void send_gpgst(const sbp2nmea_t *state) {
                sbp_pos_llh_cov->cov_n_e * sbp_pos_llh_cov->cov_n_e;
 
   /* Semi-major and semi-minor axis of the error ellipse and orientation */
-  double eigenval_1 = (trace / 2 + sqrt(((trace * trace) / 4) - det));
-  double eigenval_2 = (trace / 2 - sqrt(((trace * trace) / 4) - det));
+  double eigenval_1 = (trace / 2 + sqrt(fabs(((trace * trace) / 4) - det)));
+  double eigenval_2 = (trace / 2 - sqrt(fabs(((trace * trace) / 4) - det)));
```
akleeman: The only other option I can think of is `sqrt(max(0., ((trace * trace) / 4) - det))`, ie just flooring it at zero.  Related, if `(trace * trace / 4)` were a LOT smaller than `det` should that trigger a hard fail?

anth-cole: oh nice, much better idea.

I've not seen the second condition and theoretically it shouldn't happen, so I'm inclined not to test for it, as if we get in that situation something much much bigger went wrong upstream.

akleeman: Yep, agreed.  the `max` might mask an upstream bug ... but I think it's pretty unlikely (it'd have to be in the few lines before).

c/src/nmea.c
```diff
@@ -939,16 +939,23 @@ void send_gpgst(const sbp2nmea_t *state) {
   double det = sbp_pos_llh_cov->cov_n_n * sbp_pos_llh_cov->cov_e_e -
                sbp_pos_llh_cov->cov_n_e * sbp_pos_llh_cov->cov_n_e;
 
+  double temp;
```
akleeman: nit: could get rid of the `temp` variable in favor of a helper function,
```
double ensure_non_negative(double x);
```
then
```
`sqrt(ensure_non_negative(((trace * trace) / 4 ) - det))`
```

anth-cole: I'm going to leave this and let the compiler optimize this out

PR #205, opened 2019-05-17
Populate source field in GAL ephemeris [STAR-571]

mbavaro:
Bring in https://github.com/swift-nav/libsbp/pull/688 and https://github.com/swift-nav/libswiftnav/pull/82


c/src/rtcm3_sbp.c
```diff
@@ -249,6 +249,7 @@ void rtcm2sbp_decode_payload(const uint8_t *payload,
       if (RC_OK == rtcm3_decode_gal_eph_fnav(&payload[byte], &msg_eph)) {
         msg_ephemeris_gal_t sbp_gal_eph;
         rtcm3_gal_eph_to_sbp(&msg_eph, &sbp_gal_eph, state);
+        sbp_gal_eph.source = SBP_SOURCE_GAL_FNAV;
```
scarcanague: Nit but should we maybe pass `SBP_SOURCE_GAL_FNAV` as an argument of `rtcm3_gal_eph_to_sbp` and do the filling of `sbp_gal_eph.source` inside the function? Sounds more natural than making this post

mbavaro: sure makes sense, done

c/include/gnss-converters/rtcm3_sbp.h
```diff
@@ -43,6 +43,9 @@ extern "C" {
 #define SBP_GLO_FCN_OFFSET 8
 #define SBP_GLO_FCN_UNKNOWN 0
 
+#define SBP_SOURCE_GAL_INAV 0
+#define SBP_SOURCE_GAL_FNAV 1
```
mbavaro: I'll target https://github.com/swift-nav/libswiftnav/pull/82 and remove these

PR #207, opened 2019-05-30
Added some BDS3 signals [STAR-438]

mbavaro:
According to https://drive.google.com/file/d/1U5JvTlXC9qCj29Jl_WP5peEIoVFOGpmt/view

By the way, Allystar uses code 23 for B2a measurements, similarly to Galileo E5aQ.


mbavaro:
If we are confident that that document has the changes that ultimately will be in the standard, I could share it with Allystar and they'll do a FW update in a short time!

scarcanague:
Sounds great!

mbavaro:
Allystar confirmed that they will adjust to the provided RTCM codes, good to merge.

PR #213, opened 2019-06-12
Refactor: separate SBP3toRTCM from RTCM3toSBP

mbavaro:
RTCM to SBP and SBP to RTCM are right now completely mixed, making extending the converter capabilities a bit difficult.. this is one attempt at separating them prior to adding new functionality.
- [x] first simple shot at splitting common files in two
- [ ] collect thoughts from reviewers and apply changes if needed


mbavaro:
Yes, I had started making a GPS ephemeris encoder handler, see
https://github.com/swift-nav/gnss-converters/pull/213/files#diff-01212e3190c3b8042cece5ccee1fd8fcR54

but soon realized that something like this would be needed prior in order to keep my mind sane :)

PR #223, opened 2019-08-02
add RTCM out functionality

anth-cole:
@mfine - still really rough and needs a clean, but is now passing tests and should be the GPS eph out. There are 2 fields we need to hardcode due to SBP lacking this information - this may cause problems for 3rd party receivers


c/src/sbp_rtcm3.c
```diff
@@ -906,3 +911,20 @@ void sbp2rtcm_sbp_osr_cb(const u16 sender_id,
...
+void sbp2rtcm_sbp_gps_eph_cb(const u16 sender_id,
+                             const u8 len,
+                             const u8 msg[],
+                             struct rtcm3_out_state *state) {
+  (void)len;
+  (void)sender_id;
+  msg_ephemeris_gps_t *sbp_gps_eph = (msg_ephemeris_gps_t *)msg;
+  
+  rtcm_msg_eph msg_1019;
+  u8 frame[1024]; // Max RTCM message length is 1023 Bytes
```
mfine: there's a `RTCM3_MAX_MSG_LEN` define.

anth-cole:
sure - no problem, this should be good to go, I'll add galileo too and then can do a final clean up

anth-cole:
@mfine - think this is good to go now too

c/tests/check_rtcm3.c
```diff
@@ -825,14 +838,180 @@ static void test_SBP(const char *filename,
...
+    ck_assert_int_eq(msg_eph.kepler.w,-1872118354);
+    ck_assert_int_eq(msg_eph.kepler.omegadot,-14100);
+    ck_assert_int_eq(msg_eph.kepler.tgd_gal_s[0],-8);
+    ck_assert_int_eq(msg_eph.kepler.tgd_gal_s[1],0);
+    ck_assert_uint_eq(msg_eph.health_bits,0);
+    ck_assert_uint_eq(msg_eph.fit_interval,0);
+  }
+}
+
+static void rtcm_gal_inav_eph_cb(u8 *buffer, u16 length, void *context) {
```
martin-swift: I don't see the inav and fnav callbacks used anywhere except the unit test.  Is that intentional?

martin-swift: Ah, I see - these are here just to exercise the other functionality via test_SBP?

PR #228, opened 2019-09-09
Use absolute paths for git submodules.

benjamin0:
Is this correct?


benjaminaltieri:
@benjamin0 you may need to update your tool to allow submodules outside the root directory since the current logic will only allow `thirdparty/<submodule-repo_name>` per these lines:
https://github.com/swift-nav/ci-jenkins/blob/9fffc65e41f8860e0f6fd8b78539437b305279b5/tools/create-submodule-update.py#L112
https://github.com/swift-nav/ci-jenkins/blob/9fffc65e41f8860e0f6fd8b78539437b305279b5/tools/pyswiftnav/swift_repo.py#L104

benjamin0:
@benjaminaltieri good call, I will fix

benjamin0:
@benjaminaltieri @woodfell @anth-cole  I got it working with https urls - does this mess anything up for buildroot?

benjaminaltieri:
We will need to check things on the buildroot side but shouldn't be a big deal to fix up later

PR #237, opened 2019-09-20
Add command line options to rtcm32sbp

anth-cole:
This PR does 2 things

1. Adds the ability to specify a reference TOW (either GPS time or YYYY:MM:DD:HH) from the command line when running the rtcm32sbp binary

2. Adds the ability to handle teseoV rtcm data. the teseov sometimes outputs RTCM messages with TOW 0 before it has an internal time solve, this was causing us to stop outputting when the correct time was in the second half of the GPS week.


c/src/options.c
```diff
@@ -0,0 +1,69 @@
+/*
+ * Copyright (C) 2018 Swift Navigation Inc.
```
silverjam:
```suggestion
 * Copyright (C) 2019 Swift Navigation Inc.
```

c/src/rtcm3tosbp.c
```diff
@@ -94,12 +105,103 @@ static int read_stdin(uint8_t *buf, size_t len, void *context) {
   return read(STDIN_FILENO, buf, len);
 }
 
-int main(int argc, char **argv) {
-  (void)(argc); /* todo: accept arguments */
-  (void)(argv); /* todo: accept arguments */
+static void Help(char *arg) {
```
silverjam:
```suggestion
static void help(char *arg) {
```

c/src/rtcm3tosbp.c
```diff
@@ -17,20 +17,31 @@
...
+#include <getopt.h>
+
+#include "options.h"
 
 #define SBP_PREAMBLE 0x55
+#define STRCMP_EQ 0
+
 
 static struct rtcm3_sbp_state state;
+static void ParseBiases(char *arg);
```
silverjam: use our standard naming style

```suggestion
static void parse_biases(char *arg);
```

c/src/rtcm3tosbp.c
```diff
@@ -17,20 +17,31 @@
...
+static void ParseGlonassCodeBiases(char *arg);
```
silverjam:
```suggestion
static void parse_glonass_code_biases(char *arg);
```

c/src/rtcm3tosbp.c
```diff
@@ -17,20 +17,31 @@
...
+static void ParseGlonassPhaseBiases(char *arg);
```
silverjam:
```suggestion
static void parse_glonass_phase_biases(char *arg);
```

c/src/rtcm3tosbp.c
```diff
@@ -94,12 +105,103 @@ static int read_stdin(uint8_t *buf, size_t len, void *context) {
...
-  /* set time from systime, account for UTC<->GPS leap second difference */
+int main(int argc, char **argv) {
+  /* initialize time from systime */
   time_t ct_utc_unix = time(NULL);
+
+  int opt;
+  while ((opt = getopt(argc, argv, "hb:c:l:w:d:GRECJS:v")) != -1) {
+    switch (opt) {
+      case 'h':
+        Help(argv[0]);
```
silverjam:
```suggestion 
        help(argv[0]);
```

c/src/rtcm3tosbp.c
```diff
@@ -94,12 +105,103 @@ static int read_stdin(uint8_t *buf, size_t len, void *context) {
...
+        return 0;
+      case 'b':
+        ParseBiases(optarg);
```
silverjam:
```suggestion
        parse_biases(optarg);
```

c/src/rtcm3tosbp.c
```diff
@@ -94,12 +105,103 @@ static int read_stdin(uint8_t *buf, size_t len, void *context) {
...
+        break;
+      case 'c':
+        ParseGlonassCodeBiases(optarg);
```
silverjam:
```suggestion
        parse_glonass_code_biases(optarg);
```

c/src/rtcm3tosbp.c
```diff
@@ -94,12 +105,103 @@ static int read_stdin(uint8_t *buf, size_t len, void *context) {
...
+        break;
+      case 'l':
+        ParseGlonassPhaseBiases(optarg);
```
silverjam:
```suggestion
        parse_glonass_phase_biases(optarg);
```

c/src/rtcm3_sbp.c
```diff
@@ -1329,62 +1346,59 @@ void add_msm_obs_to_buffer(const rtcm_msm_message *new_rtcm_obs,
...
-      gpsdifftime(&obs_time, &state->last_gps_time) >= -MS_TO_S / 2) {
-    /* Find the buffer of obs to be sent */
-    msg_obs_t *sbp_obs_buffer = (msg_obs_t *)state->obs_buffer;
-
-    if (!is_msm_active(&obs_time, state) && sbp_obs_buffer->header.n_obs > 0) {
-      /* This is the first MSM observation, so clear the already decoded legacy
-       * messages from the observation buffer to avoid duplicates */
-      memset(state->obs_buffer, 0, OBS_BUFFER_SIZE);
-    }
+  /* Find the buffer of obs to be sent */
```
silverjam: the diff here is weird, but looks like the only change is the removal of the if?  any way to unit test this?

c/src/options.c
```diff
@@ -0,0 +1,69 @@
...
+#include <swiftnav/constants.h>
+#include <swiftnav/signal.h>
+
+#include "rtcm3_utils.h"
+#include "options.h"
+
+float sbp_signal_biases[CODE_COUNT] = {0.0f};
+float sbp_glo_code_bias[2] = {0.0f};
+float sbp_glo_phase_bias[2] = {0.0f};
+bool constellation_mask[CONSTELLATION_COUNT] = {false};
```
silverjam: can these be defined where they are used?  that is constellation_mask and verbosity_level

c/src/rtcm3tosbp.c
```diff
@@ -94,12 +105,103 @@ static int read_stdin(uint8_t *buf, size_t len, void *context) {
...
 
-int main(int argc, char **argv) {
-  (void)(argc); /* todo: accept arguments */
-  (void)(argv); /* todo: accept arguments */
+static void Help(char *arg) {
+  fprintf(stderr, "Usage: %s [options]\n", arg);
+  fprintf(stderr,
+          "  -h this message\n");
+  fprintf(stderr,
+          "  -b [code:bias] adds `bias` to signals with code `code`\n");
```
silverjam: typically just `-b CODE:BIAS` is used to indicate a placeholder and format

c/src/rtcm3tosbp.c
```diff
@@ -94,12 +105,103 @@ static int read_stdin(uint8_t *buf, size_t len, void *context) {
...
+          "  -l [code:bias] applies a liner phase bias to Glonass signals with code `code`s\n");
+  fprintf(stderr, 
+          "  -w [gps wn:gps tow] passes time to the converter, needs to be accurate to within a half "
+          "week for GPS only and within a half day for glonass\n");
+  fprintf(stderr, 
+          "  -d [year:month:day:hour] passes time to the nearest hour of the beginning of the data to "
+          "the converter, needs to be accurate to within a half week for GPS only and within a half day "
+          "for glonass\n");
+  fprintf(stderr, "  -[GREICJS] disables a constellation\n");
+  fprintf(stderr, "  -v [verbosity] for stderr verbosity\n");
```
silverjam: use multiple -v args to up the verbosity

c/src/rtcm3tosbp.c
```diff
@@ -120,3 +222,52 @@ int main(int argc, char **argv) {
...
   return 0;
 }
+
+static void ParseBiases(char *arg) {
+  int n,code;
+  float bias;
+  char *tok = strtok(arg, ",");
+  while (NULL != tok) {
+    n = sscanf(tok, "%d:%f", &code, &bias);
+    if ((n != 2) || (code >= CODE_COUNT)) break;
```
silverjam: error out here?

c/src/rtcm3tosbp.c
```diff
@@ -120,3 +222,52 @@ int main(int argc, char **argv) {
...
+    sbp_signal_biases[code] = bias;
+    fprintf(stderr, "Applying %.1f m bias to code %02d\n", bias, code);
```
silverjam: verbose?

c/src/rtcm3tosbp.c
```diff
@@ -120,3 +222,52 @@ int main(int argc, char **argv) {
...
+  }
+}
+
+static void ParseGlonassPhaseBiases(char *arg) {
+  int n, code;
+  float bias_fcn_ratio;
+  char *tok = strtok(arg, ",");
+  while (NULL != tok) {
+    n = sscanf(tok, "%d:%f", &code, &bias_fcn_ratio);
+    if (n != 2) break;
```
silverjam: since all of these follow a similar pattern, same Qs about error handling and verbosity

silverjam:
was the libsbp pointer change intentional?

c/src/rtcm3_sbp.c
anth-cole: This one I'll split it out into a separate commit. Essentially, I _believe_ this if to be redundant, in that it checks if the next obs is somewhere from [-0.5s,infinity) time gap from the previously broadcast message. Removing this still passes all our unit tests, which makes me much more comfortable that it still behaves correctly. Tracing the git history, this was originally trying to ensure we don't send 2 SBP messages for both legacy and MSM inputs, but this is now handled using a different mechanism.

anth-cole: The reason this was causing problems was that the TeseoV will output a series of observation messages with TOW set to 0 before it resolves the correct time internally. If the correct time is in the second half of the week, we then adjust the TOW into the past, causing us to fail this check and not send out valid SBP messages

anth-cole:
@silverjam I think I addressed all your feedback, thanks! The only one I couldn't see a better option for was the location of the 'verbosity' and 'constellation mask' variables. When I tried to add them to rtcm32sbp.c I got linking errors I haven't been able to figure out why.

I split the PR into 3 commits
1. Add command line options
2. Revert the libsbp pointer update
3. Adds the functionality to deal with TeseoV data that seems to send messages with TOW = 0 whenever it doesn't have an internal time solve. This was causing problems in a lot of the time syncronization code so in essence I've filtered these epochs out. This may cost us 1 observation a week but I don't have a better solution right now

anth-cole:
@silverjam yeah - in the process of adding tests now, will get the new code covered and get that green before I land

anth-cole:
going to land this. Added a unit test, but the degredation in unit test coverage is due to the following changes
- Adding static helper functions to the rtcm3tosbp.c file that are helper functions to parse command line arguments
- Additional logging and support of logging levels mean we don't hit some logging fprintf statements in the code however turning this on for unit testing makes the tests very noisy.

silverjam:
@anth-cole to deal with this in the future we should probably have more mechanisms to exclude or disable code that isn’t appropriate for unit testing

silverjam:
@anth-cole besides excluding code, the other option would be to mock things like printf to avoid spamming the test output (mocking could also be a mechanism for excluding code from the unit tests)

PR #240, opened 2019-09-26
libswiftnav to master.  Should be automated.

martin-swift:


martin-swift:
https://github.com/swift-nav/orion/pull/786#issuecomment-535681109

PR #241, opened 2019-10-04
move options.h to include/gnss-converters [ESD-1565]

benjaminaltieri:
this will allow the headers to be installed properly as currently this causes a failure on any downstream repo that tries to install gnss-converters


denniszollo:
Sorry I missed the review last night; thanks for fixing Ben!

PR #244, opened 2019-10-11
Add ubx support [STAR-840]

yizow:


c/src/ubx_sbp.c
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    ret = read_ubx_bytes(frame, 4, state);
+    if (ret == -1) {
+      return -1;
+    }
+
+    u8 class_id = frame[0];
+    u8 msg_id = frame[1];
+    (void)class_id;
```
anth-cole: can remove these if we're not using them

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,367 @@
...
+  if (frac_part_cp >= UINT8_MAX) {
+    frac_part_cp = 0;
+    L_out->i += 1;
+  }
+  L_out->f = frac_part_cp;
+
+  return true;
+}
+
+bool pack_doppler(double D_in, doppler_t *D_out) {
```
anth-cole: Do these functions exist in libsbp/libswiftnav?

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,367 @@
...
+ */
+int ubx_sbp_process(struct ubx_sbp_state *state) {
+  u8 inbuf[UBX_FRAME_SIZE];
+
+  ssize_t ret = read_ubx_frame(inbuf, state);
+  if (ret <= 0) {
+    return ret;
+  }
+
+  if ((inbuf[0] == 0x01 && inbuf[1] == 0x07)) {
```
anth-cole: Can we use defines here? not sure what 0x01 and 0x07 are but assume class/message ID? might be better to define them in libubx

anth-cole: maybe even a function in libubx is_'message'?

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,367 @@
...
+    msg_pos_llh_t sbp_pos_llh;
+    if (fill_msg_pos_llh(inbuf, &sbp_pos_llh) == 0) {
+      uint16_t sbp_msg_type = SBP_MSG_POS_LLH;
+      state->write_func(
+          (u8 *)&sbp_pos_llh, sizeof(sbp_pos_llh), (void *)&sbp_msg_type);
+    }
+  }
+
+  if ((inbuf[0] == 0x02 && inbuf[1] == 0x15)) {
+    u8 sbp_obs_buffer[sizeof(msg_obs_t) +
```
anth-cole: this looks like it could be broken out into a helper function once we've detected an obs message

c/tests/check_ubx.h
```diff
@@ -0,0 +1,27 @@
...
+ * This source is subject to the license found in the file 'LICENSE' which must
+ * be be distributed together with this source. All other rights reserved.
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef CHECK_UBX_H
+#define CHECK_UBX_H
```
anth-cole: Not overly opposed, but do we need this header? Seems it's only defining the MAX_FILE_SIZE?

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,367 @@
...
+      msg->obs[i].cn0 = 0xFF;
+    } else {
+      msg->obs[i].cn0 = 4 * rxm_rawx.cno_dbhz[i];
+    }
+    // TODO(yizhe) converts from u32 ms, to double s; encode_lock_time
+    // internally converts back to u32 ms.
+    msg->obs[i].lock = encode_lock_time((double)rxm_rawx.lock_time[i] / 1000.0);
+    msg->obs[i].flags = 0;
+    // currently assumes all doppler are valid
+    msg->obs[i].flags |= 1 << 3;
```
jbangelo: I'm torn about all these magic numbers. On the one hand they're all magic numbers and our guidelines dictate that they should be put into named constants. On the other hand they're all basically used in one place and are just conversion coefficients so that name would be something like `LOCK_TIME_CONVERSION_COEFF`.

I think I'd prefer to see these as named constants.

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,367 @@
...
+}
+
+int fill_msg_pos_llh(const u8 buf[], msg_pos_llh_t *msg) {
+  ubx_nav_pvt nav_pvt;
+  if (ubx_decode_nav_pvt(buf, &nav_pvt) != RC_OK) {
+    return -1;
+  }
+
+  msg->tow = nav_pvt.i_tow;
+  double scaling_factor = 1e-7;
```
jbangelo:
```suggestion
  const double scaling_factor = 1e-7;
```
?

c/src/ubx_sbp.c
jbangelo: Agreed.

c/src/ubx_sbp.c
martin-swift: maybe even `constexpr`?

jbangelo: Can't use `constexpr`, this is a C file.

c/include/gnss-converters/ubx_sbp.h
```diff
@@ -0,0 +1,55 @@
...
+ *
+ * THIS CODE AND INFORMATION IS PROVIDED "AS IS" WITHOUT WARRANTY OF ANY KIND,
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#ifndef GNSS_CONVERTERS_UBX_SBP_INTERFACE_H
+#define GNSS_CONVERTERS_UBX_SBP_INTERFACE_H
+
+#include <stdio.h>
```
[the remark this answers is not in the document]
yizow: Moved the `ubx2sbp.c`

c/src/ubx2sbp.c
```diff
@@ -0,0 +1,42 @@
...
+  (void)argc;
+  (void)argv;
+  // TODO(yizhe) accept sender id as a cmdline argument
+
+  sbp_state_t sbp_state;
+  sbp_state_init(&sbp_state);
+
+  struct ubx_sbp_state state;
+  state.read_stream_func = &read_stdin;
+  reload_ubx_sbp_state(&state);
```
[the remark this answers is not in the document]
yizow: This function reloads the read buffer.
Initialization would be literally just a wrapper around this function, but named `init` instead. It felt off for me to name this `init`, but then call it repeatedly. as the read buffer runs out

Ah, seeing the comment about setting read/write funcs below makes more sense. I'll add in an `init` to handle all that.

c/tests/check_ubx.h
yizow: Yes, that's all its doing right now.
I added it in for a couple reasons:
1) We've only implemented minimal ubx support. I can imagine this requiring more for other messages, or more extensive tests.
2) to mirror existing rtcm format

c/src/ubx_sbp.c
yizow: Yep, this code was based off of lines in `starling`. Didn't want to introduce that circular dependency, so re-implemented them here.
https://github.com/swift-nav/starling/blob/master/src/util/sbp/packers.c#L418-L480

I can add a TODO + jira ticket to move this into libsbp

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,367 @@
...
+#include <swiftnav/signal.h>
+
+#include <ubx/decode.h>
+#include <ubx/ubx_messages.h>
+
+// TODO(yizhe) should probably consolidate these into central .h file
+#define MSG_OBS_LF_MULTIPLIER 256
+#define MSG_OBS_DF_MULTIPLIER 256
+
+// floor((0xFF - sizeof(observation_header_t)) / sizeof(packed_obs_content_t))
```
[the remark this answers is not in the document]
yizow: Done

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,367 @@
...
+
+#include <gnss-converters/ubx_sbp.h>
+
+#include <swiftnav/nav_meas.h>
+#include <swiftnav/signal.h>
+
+#include <ubx/decode.h>
+#include <ubx/ubx_messages.h>
+
+// TODO(yizhe) should probably consolidate these into central .h file
```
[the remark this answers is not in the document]
yizow: Done

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,367 @@
...
+#define SBP_MAX_NUM_OBS 14
+
+void reload_ubx_sbp_state(struct ubx_sbp_state *state) {
+  memset(state->read_buffer, 0, UBX_BUFFER_SIZE);
+  state->index = 0;
+  state->bytes_in_buffer =
+      state->read_stream_func(state->read_buffer, UBX_BUFFER_SIZE, NULL);
+}
+
+// Attempts to read `length` bytes into `buf`, implemented as a memcpy from
```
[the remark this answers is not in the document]
yizow: All Fixed

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,367 @@
...
+
+int fill_msg_obs(const u8 buf[], msg_obs_t *msg) {
+  ubx_rxm_rawx rxm_rawx;
+  if (ubx_decode_rxm_rawx(buf, &rxm_rawx) != RC_OK) {
+    return -1;
+  }
+
+  // convert from sec to ms
+  double ms = 1e3 * rxm_rawx.rcv_tow;
+  double ns = 1e6 * modf(ms, &ms);
```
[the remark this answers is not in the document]
yizow: Not explicitly allowed, but i also couldn't find anything that prohibited it.
https://stackoverflow.com/a/56223182

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,367 @@
...
+    if (state->bytes_in_buffer == 0) {
+      // EOF reached
+      return -1;
+    }
+
+    size_t remaining_bytes = state->bytes_in_buffer - state->index;
+    memcpy(buf, &state->read_buffer[state->index], remaining_bytes);
+    length -= remaining_bytes;
+    buf += remaining_bytes;
+    reload_ubx_sbp_state(state);
```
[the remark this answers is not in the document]
yizow: This loop condition triggers if we are attempting to read more bytes than the buffer currently has.
First we read out the remaining bytes in buffer, then reload the buffer and recheck condition.

c/src/ubx_sbp.c
yizow: Moved into a `#define`

yizow:
code-cov decrease is due to error cases and various signal ID combinations, not "real" code coverage decreases.

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,409 @@
...
+  double ns = UBX_SBP_TOW_NS_SCALING * modf(ms, &ms);
+  msg->header.t.tow = (u32)ms;
+  msg->header.t.ns_residual = (s32)ns;
+  if (msg->header.t.ns_residual > 500000) {
+    msg->header.t.ns_residual -= 1000000;
+    msg->header.t.tow += 1;
+  }
+
+  msg->header.t.wn = rxm_rawx.rcv_wn;
+  msg->header.n_obs = rxm_rawx.num_meas;
```
jbangelo: This isn't how the `n_obs` member is structured.

Also are we guaranteed that a single UBX obs message can fit into a single SBP obs message?

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,409 @@
...
+    return -1;
+  }
+
+  msg->tow = nav_pvt.i_tow;
+  msg->lat = UBX_SBP_LAT_LON_SCALING * nav_pvt.lat;
+  msg->lon = UBX_SBP_LAT_LON_SCALING * nav_pvt.lon;
+  /* convert from mm to m */
+  msg->height = UBX_SBP_HEIGHT_SCALING * nav_pvt.height;
+  /* bounding u32 -> u16 conversion */
+  u16 max_accuracy = -1;
```
jbangelo: This isn't obvious what you're doing. We should be using `UINT16_MAX` here instead.

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,409 @@
...
+      for (obs_index = 0;
+           obs_index < SBP_MAX_NUM_OBS &&
+           (obs_index + msg_num * SBP_MAX_NUM_OBS) < sbp_obs->header.n_obs;
+           obs_index++) {
+        sbp_obs_to_send->obs[obs_index] =
+            sbp_obs->obs[obs_index + msg_num * SBP_MAX_NUM_OBS];
+      }
+
+      u32 total_msg_size = sizeof(observation_header_t) +
+                           sizeof(packed_obs_content_t) * obs_index;
```
jbangelo: I don't think `obs_index` is correct here. Isn't it accumulating the total number of obs, when it should be just the number of obs in this message?

c/include/gnss-converters/ubx_sbp.h
```diff
@@ -0,0 +1,49 @@
...
+struct ubx_sbp_state {
+  u8 read_buffer[UBX_BUFFER_SIZE];
+  size_t index;
+  size_t bytes_in_buffer;
+  int (*read_stream_func)(u8 *buf, size_t len, void *ctx);
+  s32 (*cb_ubx_to_sbp)(u8 *buf, u32 len, void *ctx);
+};
+
+void ubx_sbp_init(struct ubx_sbp_state *state,
+                  s32 (*cb_ubx_to_sbp)(u8 *buff, u32 length, void *context));
```
jbangelo: I think it would make this more useful if it split out the components of the SBP message as arguments like the RTCM conversion does. `msg_id`, `length`, `buffer`. We could probably drop the sender ID, but should make it clear why one isn't provided.

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,409 @@
...
+  if (class_id == UBX_CLASS_RXM && msg_id == UBX_MSG_RXM_RAWX) {
+    u8 sbp_obs_buffer[sizeof(msg_obs_t) +
+                      sizeof(packed_obs_content_t) * UBX_MAX_NUM_OBS];
+    handle_rxm_rawx(state, frame, sbp_obs_buffer);
+  }
+}
+
+void ubx_sbp_init(struct ubx_sbp_state *state,
+                  s32 (*cb_ubx_to_sbp)(u8 *buff, u32 length, void *context)) {
+  memset(state->read_buffer, 0, UBX_BUFFER_SIZE);
```
jbangelo: Do we really need to zero out the memory?

c/tests/check_ubx.c
```diff
@@ -0,0 +1,115 @@
...
+  uint8_t tmpbuf[5];
+  tmpbuf[0] = (uint8_t)sbp_msg_type;
+  tmpbuf[1] = (uint8_t)(sbp_msg_type >> 8);
+  tmpbuf[2] = (uint8_t)DEFAULT_UBX_SENDER_ID;
+  tmpbuf[3] = (uint8_t)(DEFAULT_UBX_SENDER_ID >> 8);
+  tmpbuf[4] = (uint8_t)n;
+
+  u16 crc = crc16_ccitt(tmpbuf, sizeof(tmpbuf), 0);
+  crc = crc16_ccitt(buff, n, crc);
+  ck_assert(crc == rxm_rawx_crc[msg_index]);
```
jbangelo: It doesn't look like we're checking the values in the message?

yizow: I'm assuming if the calculated crc matches with a crc of a message with the correct values, the messages match.

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,409 @@
...
+#define SBP_OBS_DOPPLER_MASK (1 << 3)
+#define SBP_OBS_TRACK_STATE_MASK (0x07)
+
+/* A single SBP message can only fit a maximum number of observations in it
+ * floor((0xFF - sizeof(observation_header_t)) / sizeof(packed_obs_content_t))
+ */
+#define SBP_MAX_NUM_OBS 14
+
+static void reload_ubx_sbp_state(struct ubx_sbp_state *state) {
+  memset(state->read_buffer, 0, UBX_BUFFER_SIZE);
```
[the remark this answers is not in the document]
yizow: I guess this was just habit.
you're right, its not actually doing anything here.

c/src/ubx_sbp.c
yizow: I'm using `n_obs` akin to a temporary variable here; its reformatted between being sent.
It's confusing as currently written; i'll use an actual temp variable for clarity

yizow: No, we do not have that guarantee. Their max size is `UINT8_MAX`, while ours is `UINT4_MAX`

c/src/ubx_sbp.c
yizow: The tricky thing here is that UBX's `num_meas` and SBP's `n_obs` are not just different types (one's a uint8, ones two uint4 (total num message, index)), but the thing theyre counting is also different.

UBX `num_meas` counts the number of actual observations
SBP `n_obs` counts the number of SBP messages needed to send all the actual observations.

SBP does not explicitly send the number of obs in a message. It relies on the receiver to calculate it from the length of the message.

yizow: Wait sorry, i think i misread your comment

yizow: `obs_index` is set to 0 by the inner for loop at the start of each message. Should be accumulating number of obs per message, not total number of obs.

c/tests/check_ubx.c
jbangelo: Oh, I think I'm following now. All the messages contain the same data and we're just comparing the CRC against a precomputed checksum.

yizow: Yep. the expected output is 2 SBP messages; CRC values here:
https://github.com/swift-nav/gnss-converters/pull/244/files/8dc9e6a9f4ebd926112baf0e90d9ec61ccdf9877#diff-03034945e5ed17eecf3c576f7aabb527R30

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,422 @@
...
+      state->cb_ubx_to_sbp(SBP_MSG_OBS,
+                           (u8)total_msg_size,
+                           (u8 *)sbp_obs_to_send,
+                           DEFAULT_UBX_SENDER_ID,
+                           state->context);
+    }
+  }
+}
+
+void ubx_handle_frame(u8 *frame, struct ubx_sbp_state *state) {
```
jbangelo: Might as well declare this in the header file.

c/src/ubx_sbp.c
```diff
@@ -0,0 +1,422 @@
...
+  return 0;
+}
+
+static void handle_nav_pvt(struct ubx_sbp_state *state, u8 *inbuf) {
+  msg_pos_llh_t sbp_pos_llh;
+  if (fill_msg_pos_llh(inbuf, &sbp_pos_llh) == 0) {
+    state->cb_ubx_to_sbp(SBP_MSG_POS_LLH,
+                         sizeof(sbp_pos_llh),
+                         (u8 *)&sbp_pos_llh,
+                         DEFAULT_UBX_SENDER_ID,
```
jbangelo: Could we make this configurable? I think defaulting it's value to something constant like this is fine, but I could see a need to change it to something different in the future. It would just need to be another function declared in the header.

PR #245, opened 2019-10-15
clang-format-all fixes

yizow:
Changes are entirely from running `make clang-format-all`.


yizow:
Looks like Codecov check is failing because it's format splitting un-code-cov'd lines into two, thus increasing the number of uncovered lines...

Since this is a purely formatting PR, can this be one of those times we overrule the codecov rule?

martin-swift:
> Looks like Codecov check is failing because it's format splitting un-code-cov'd lines into two, thus increasing the number of uncovered lines...
> 
> Since this is a purely formatting PR, can this be one of those times we overrule the codecov rule?

Absolutely - that is exactly the time to ignore CC.

PR #247, opened 2019-10-17
Add licence file

anth-cole:
Adding GPLv3 license file


silverjam:
Close?  Looks like we already have a license?

silverjam:
LICENSE file is already there (and it's already GPLv3)-- re-open if needed

PR #248, opened 2019-10-21
Use gps_time_match_weeks() on ephemeris

mfine:
This was intended to be the code path before. Bring it back in.


mfine:
Need to make the types work out, but the general idea here is to call this on every ephemeris.

c/src/rtcm3_sbp_ephemeris.c
```diff
@@ -335,6 +343,7 @@ void rtcm3_bds_eph_to_sbp(rtcm_msg_eph *msg_eph,
   sbp_bds_eph->common.toe.wn =
       gps_adjust_week_cycle(toe.wn, GPS_WEEK_REFERENCE);
   sbp_bds_eph->common.toe.tow = (u32)rint(toe.tow);
+  gps_time_sec_match_weeks(&sbp_bds_eph->common.toe, &state->time_from_rover_obs);
```
pgrgich: I gather `toe` refers to both the clock and orbit time in this case, does it?

mfine: Good point - I wanted to only handle GPS and GAL for now. Can revisit BDS later.

mfine:
@pgrgich I didn't find any hard evidence outside of what was posted in comments and RINEX docs. I forgot to link the ticket: https://swift-nav.atlassian.net/browse/ORI-776.

PR #275, opened 2019-11-08
Pass context in read, check for negative err codes

yizow:


c/src/ubx_sbp.c
```diff
@@ -39,8 +39,8 @@
 
 static void reload_ubx_buffer(struct ubx_sbp_state *state) {
   state->index = 0;
-  state->bytes_in_buffer =
-      state->read_stream_func(state->read_buffer, UBX_BUFFER_SIZE, NULL);
+  int bytes_read = state->read_stream_func(state->read_buffer, UBX_BUFFER_SIZE, state->context);
+  state->bytes_in_buffer = bytes_read > 0 ? bytes_read : 0;
```
denniszollo: I am thinking this one might need a return code to pass up the errors stat from the reading interface.

yizow: That's a good point, fixing

denniszollo:
I'm not really sure all the implications of this; but if you are confident it is an improvement I have no objections.  In general, I feel like we are taking a bunch of possible error states and turning them into the "0 bytes read" case. 

I worry that there are situations (especially on serial) where people might use read functions that don't return immediately if there is an empty buffer, which is a normal non/error state that shouldn't result in closing the ubx reader. 

Lets just keep our eye on this when we start integrating libubx into real time starling runner.

yizow:
You're correct about us masking potential error states. This PR was born out of starling runner integration efforts and the various edge cases around serial, file I/O, etc. 
I've pushed up a new commit that should pass the error code all the way up to the calling application.

PR #276, opened 2019-11-13
Correctly convert ubx nav_pvt to sbp llh fix mode [ESD-1624]

yizow:


c/src/ubx_sbp.c
```diff
@@ -311,17 +327,30 @@ static int fill_msg_pos_llh(const u8 buf[], msg_pos_llh_t *msg) {
...
+    if ((nav_pvt.flags & UBX_NAV_PVT_DGNSS_MASK) == 0) {
       msg->flags |= SBP_LLH_SPP_MASK;
-    }
-    /* check for INS. */
-    if (nav_pvt.flags == UBX_NAV_PVT_INS_MASK) {
-      msg->flags |= SBP_LLH_INS_MASK;
+    } else {
+      u8 carr_soln = nav_pvt.flags & UBX_NAV_PVT_CARR_SOLN_MASK;
+      switch (carr_soln) {
+        case UBX_NAV_PVT_CARR_SOLN_NONE:
```
jbangelo: What does this solution type mean? `SOLN_NONE` makes it sound like an invalid solution from the outside.

c/src/ubx_sbp.c
```diff
@@ -311,17 +327,30 @@ static int fill_msg_pos_llh(const u8 buf[], msg_pos_llh_t *msg) {
...
+          msg->flags |= SBP_LLH_DGNSS_MASK;
+          break;
+        case UBX_NAV_PVT_CARR_SOLN_FLOAT:
+          msg->flags |= SBP_LLH_FLOAT_RTK_MASK;
+          break;
+        case UBX_NAV_PVT_CARR_SOLN_FIXED:
+          msg->flags |= SBP_LLH_FIXED_RTK_MASK;
+          break;
+        default:
```
jbangelo: Do we have a way to log error messages? Or can this function return an error code? We should probably indicate an error in some way if we run into a solution type we don't understand.

c/tests/check_ubx.c
```diff
@@ -97,6 +97,48 @@ static void ubx_sbp_callback_nav_pvt_corrupted(
   msg_index++;
 }
 
+uint16_t nav_pvt_fix_type_crc[] = {32103,
```
jbangelo: `static const` this?

c/src/ubx_sbp.c
yizow:
```
Carrier phase range solution status:
0: no carrier phase range solution
```

Wasn't sure how to phrase it.

c/src/ubx_sbp.c
yizow: Nothing more than a `fprintf(stderr)`. I'll add it in

c/src/ubx_sbp.c
```diff
@@ -17,9 +17,25 @@
...
+
 #define UBX_NAV_PVT_DGNSS_MASK (0x02)
 #define UBX_NAV_PVT_INS_MASK (0x04)
 
+#define UBX_NAV_PVT_CARR_SOLN_MASK (0xC0)
+#define UBX_NAV_PVT_CARR_SOLN_NONE (0x00)
+#define UBX_NAV_PVT_CARR_SOLN_FLOAT (0x40)
+#define UBX_NAV_PVT_CARR_SOLN_FIXED (0xC0)
+
+#define SBP_LLH_INVALID_RTK_MASK (0x00)
```
jbangelo: This isn't Invalid RTK per se, it's just an invalid solution.

c/src/ubx_sbp.c
```diff
@@ -311,17 +327,30 @@ static int fill_msg_pos_llh(const u8 buf[], msg_pos_llh_t *msg) {
...
-  /* if valid fix */
-  if (nav_pvt.flags & UBX_NAV_PVT_DGNSS_MASK) {
-    /* if differential corrections */
-    if (nav_pvt.flags & UBX_NAV_PVT_DGNSS_MASK) {
-      msg->flags |= SBP_LLH_DGNSS_MASK;
-    } else {
+  if (nav_pvt.fix_type == UBX_NAV_PVT_FIX_TYPE_NONE) {
+    msg->flags |= SBP_LLH_INVALID_RTK_MASK;
+  } else if (nav_pvt.fix_type == UBX_NAV_PVT_FIX_TYPE_DEAD_RECKONING) {
+    msg->flags |= SBP_LLH_DEAD_RECKONING_MASK;
```
jbangelo: We should probably also set the SBP INS used flag here. Ideally we'd set it in the other situations where it makes sense, but I guess UBX doesn't provide that info?

c/src/ubx_sbp.c
```diff
@@ -311,17 +327,30 @@ static int fill_msg_pos_llh(const u8 buf[], msg_pos_llh_t *msg) {
...
+  } else if ((nav_pvt.fix_type == UBX_NAV_PVT_FIX_TYPE_2D) ||
+             (nav_pvt.fix_type == UBX_NAV_PVT_FIX_TYPE_3D) ||
+             (nav_pvt.fix_type == UBX_NAV_PVT_FIX_TYPE_COMBINED)) {
+    if ((nav_pvt.flags & UBX_NAV_PVT_DGNSS_MASK) == 0) {
```
jbangelo: Wouldn't it make sense to set the SBP DGNSS mode here instead of SPP? Or is this how UBX indicates a SPP solution?

c/src/ubx_sbp.c
yizow: `INVALID_RTK`->`INVALID_SOLN`

c/src/ubx_sbp.c
```diff
@@ -311,17 +327,30 @@ static int fill_msg_pos_llh(const u8 buf[], msg_pos_llh_t *msg) {
   msg->n_sats = nav_pvt.num_sats;
 
   msg->flags = 0;
-  /* if valid fix */
-  if (nav_pvt.flags & UBX_NAV_PVT_DGNSS_MASK) {
-    /* if differential corrections */
-    if (nav_pvt.flags & UBX_NAV_PVT_DGNSS_MASK) {
-      msg->flags |= SBP_LLH_DGNSS_MASK;
-    } else {
+  if (nav_pvt.fix_type == UBX_NAV_PVT_FIX_TYPE_NONE) {
```
jbangelo: Should also probably catch the "time only fix" type here.

c/src/ubx_sbp.c
jbangelo: Ok looks like they sorta kinda give than info. Dead reckoning should definitely set the SBP INS used flag. We should also set it if the fix mode is "GNSS + dead reckoning combined"

c/src/ubx_sbp.c
jbangelo: Never mind, I see this is for when the DGNSS flag is not set. Ooops 😳

c/src/ubx_sbp.c
yizow: Just added that in

c/src/ubx_sbp.c
jbangelo: Maybe
```UBX_NAV_PVT_SOLN_NO_PHASE
UBX_NAV_PVT_SOLN_FLOAT_RTK
UBX_NAV_PVT_SOLN_FIXED_RTK
```
?

Not really a blocker for this PR, but would make it more clear I feel.

c/src/ubx_sbp.c
```diff
@@ -311,18 +327,40 @@ static int fill_msg_pos_llh(const u8 buf[], msg_pos_llh_t *msg) {
...
+          msg->flags |= SBP_LLH_DGNSS_MASK;
+          break;
+        case UBX_NAV_PVT_SOLN_FLOAT_RTK:
+          msg->flags |= SBP_LLH_FLOAT_RTK_MASK;
+          break;
+        case UBX_NAV_PVT_SOLN_FIXED_RTK:
+          msg->flags |= SBP_LLH_FIXED_RTK_MASK;
+          break;
+        default:
+          fprintf(stderr,
```
jbangelo: Do we log to `stderr` anywhere else in this library? It's not usually a great thing to do since the application might be using that for a specific purpose, or it might not go anywhere useful. Would a return code be more appropriate to indicate that the function encountered a situation it couldn't handle?

yizow: yep, there's `fprintf(stderr,` all over rtcm2sbp

PR #281, opened 2019-11-19
fix typo E5{I/Q} -> E7{I/Q} [STAR-969]

yizow:


yizow:
Sample dataset only has `E7Q` obs, not `E7I`

PR #297, opened 2019-12-19
Yizow/ubx nav velecef

yizow:


c/src/ubx_sbp.c
```diff
@@ -382,6 +385,31 @@ static int fill_msg_pos_llh(const u8 buf[], msg_pos_llh_t *msg) {
...
+}
+
+static int fill_msg_vel_ecef(const u8 buf[], msg_vel_ecef_t *msg) {
+  ubx_nav_velecef nav_velecef;
+  if (ubx_decode_nav_velecef(buf, &nav_velecef) != RC_OK) {
+    return -1;
+  }
+
+  msg->tow = nav_velecef.i_tow;
+  msg->x = nav_velecef.ecefVX * UBX_NAV_VELECEF_SCALING;
```
anth-cole: nit: rather than make this a specific UBX velocity scaling define, it can be more reusable to define a general CM_TO_MM define that can be used globally for those conversions.

yizow: There's a host of constants in this file that I'd prefer to move to a different global place, but havent had a chance to get to it yet
https://swift-nav.atlassian.net/browse/ESD-1625

PR #311, opened 2020-01-23
Added GPS L1CA UBX ephemeris decoder

adel-mamin:
The PR adds GPS L1CA ephemeris data codec for the conversion from Ublox UBX format to SwiftNav format.

The Ublox format: https://www.u-blox.com/sites/default/files/products/documents/u-blox8-M8_ReceiverDescrProtSpec_(UBX-13003221)_Public.pdf

The PR was tested by running Ublox EVK-M8U and SwiftNav Duro with the same signal, processing the resulting logs and comparing the converted Ublox GPS L1CA ephemeris data against Duro's ephemeris data.

The PR is contingent on https://github.com/swift-nav/libubx/pull/15


jbangelo:
This appears to be decoding Ublox data? If that's the case it seems like something that should live in https://github.com/swift-nav/libubx

c/src/ubx_ephe/gps_l1ca.c
```diff
@@ -0,0 +1,167 @@
...
+}
+
+/**
+ * Decodes GPS L1CA subframes.
+ * Reference: ICD IS-GPS-200H
+ * @param prn transmitter's PRN
+ * @param subframe the array of full subframe (30 bit words)
+ * @param sz must be 10 (number of 30 bit words per one GPS L1CA subframe)
+ */
+void gps_l1ca_decode_subframe(int prn, const u32 words[], int sz) {
```
jbangelo: I think it would be better if we split up this logic into ublox deframing and decoding the raw ephemeris data into an SBP format.

I'm thinking that Libubx would define a new class/message type for ephemeris, and the message payload would consist of the array of `struct sat`s. Then in gnss-converters we would use libswiftnav to decode this data into SBP messages.

Does that make sense?

adel-mamin: > Does that make sense?

Let's see. 
The ublox `UBX-RXM-SFRBX` de-framing in the PR is done with `https://github.com/swift-nav/libubx/blob/master/c/src/decode.c#L593`. The resulting structure is described here:
https://github.com/swift-nav/libubx/blob/master/c/include/ubx/ubx_messages.h#L218-L231
It contains one raw GPS L1CA subframe in `ubx_rxm_sfrbx::data_words[10]` array. 
`gps_l1ca_decode_subframe()` combines raw subframes 1,2 and 3 are handles them to libswiftnav's `decode_ephemeris()`.

> I'm thinking that Libubx would define a new class/message type for ephemeris, and the message payload would consist of the array of struct sats.

We could do this although then we would need to convert it back to raw GPS L1CA subframes 1,2 and 3 in gnss-converters code to use libswiftnav `decode_ephemeris()` function.

jbangelo: Ok, I think I see how this works now. Let's disregard me previous comment. If the Ublox message gives us one subframe at a time and libswiftnav's ephemeris decoding requires subframes 1-3 then I think it makes sense to keep the buffering of the subframes here.

c/src/ubx_ephe/gps_l1ca.c
```diff
@@ -0,0 +1,167 @@
...
+
+struct gps_l1ca_data {
+  struct sat sat[NUM_SATS_GPS];
+  void *context;
+  void (*cb_ubx_to_sbp)(
+      u16 msg_id, u8 length, u8 *buff, u16 sender_id, void *context);
+  u16 sender_id;
+};
+
+static struct gps_l1ca_data m_data;
```
jbangelo: Let's not have this as a static global variable. That prevents us from being able to run multiple instances of the UBX to SBP conversion in a single process. Could we move this to be in the `ubx_sbp_state` context struct? That would remove the need for the callback, context, and sender_id fields since those are already present in the `ubx_sbp_state` struct.

c/src/ubx_ephe/gps_l1ca.c
```diff
@@ -0,0 +1,167 @@
...
+  msg->af0 = (float)k->af0;
+  msg->af1 = (float)k->af1;
+  msg->af2 = (float)k->af2;
+  msg->toc.tow = (u32)round(k->toc.tow);
+  msg->toc.wn = k->toc.wn;
+  msg->iode = k->iode;
+  msg->iodc = k->iodc;
+}
+
+void gps_l1ca_init(void (*cb_ubx_to_sbp)(u16 msg_id,
```
jbangelo: I don't think we need a separate callback registration for the ephemeris. Couldn't we simply use the callback in the `ubx_sbp_state` struct?

c/src/CMakeLists.txt
```diff
@@ -7,10 +7,28 @@ set(gnss_converters_HEADERS
...
 
-add_library(gnss_converters rtcm3_sbp.c sbp_rtcm3.c rtcm3_sbp_ephemeris.c rtcm3_sbp_ssr.c ubx_sbp.c sbp_nmea.c nmea.c rtcm3_utils.c sbp_conv.c sbp_rtcm3_ephemeris.c options.c)
+add_library(
+  gnss_converters
+  rtcm3_sbp.c
+  sbp_rtcm3.c
+  rtcm3_sbp_ephemeris.c
+  rtcm3_sbp_ssr.c
+  ubx_sbp.c
+  ${PROJECT_SOURCE_DIR}/src/ubx_ephe/ubx_ephe.c
```
yizow: Minor nit:
Can we just type out the whole `ephemeris`? Maybe this is my lack of GNSS domain knowledge, but I don't think `ephe` is a common abbreviation.

c/src/ubx_sbp.c
```diff
@@ -1,17 +1,16 @@
+#include <gnss-converters/ubx_sbp.h>
```
yizow: Is there a reason to shuffle these import statements around? Previously, they were grouped by what "library" they were part of.

Under that schema, `#include "ubx_epehe.h"` would be `#include <gnss-converters/ubx_ephe.h>`

c/src/ubx_sbp.c
```diff
@@ -591,6 +590,8 @@ void ubx_handle_frame(u8 *frame, struct ubx_sbp_state *state) {
         u8 sbp_obs_buffer[sizeof(msg_obs_t) +
                           sizeof(packed_obs_content_t) * UBX_MAX_NUM_OBS];
         handle_rxm_rawx(state, frame, sbp_obs_buffer);
+      } else if (msg_id == UBX_MSG_RXM_SFRBX) {
+        ubx_ephe_handle_rxm_sfrbx(state, frame, UBX_FRAME_SIZE);
```
yizow: Regarding Joseph's comments earlier about using `ubx_sbp_state` to avoid duplication, it'd look like defining a `static void handle_rxm_sfrbx()` function in this file, changing this to call that. 

It'd allow us to remove the `ubx_ephe_init` line later as well.

yizow: I think it'd also allow us to remove `ubx_ephe.c`, and just have `ubx_ephe.h` header hold the `#define` constants.

c/src/ubx_ephe/ubx_ephe.c
```diff
@@ -0,0 +1,44 @@
...
+ * EITHER EXPRESSED OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE IMPLIED
+ * WARRANTIES OF MERCHANTABILITY AND/OR FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#include "gnss-converters/ubx_sbp.h"
+#include "gps_l1ca.h"
+#include "swiftnav/common.h"
+#include "ubx/decode.h"
+
+#define UBX_GNSS_ID_GPS 0
```
yizow: Actually, would these make sense to go in `libubx`, since they are ubx specific constants?
https://github.com/swift-nav/libubx/blob/master/c/include/ubx/constants.h

c/tests/check_ubx.c
```diff
@@ -101,6 +101,28 @@ static void ubx_sbp_callback_rxm_rawx(
...
+  uint8_t tmpbuf[5];
+  tmpbuf[0] = (uint8_t)msg_id;
+  tmpbuf[1] = (uint8_t)(msg_id >> 8);
+  tmpbuf[2] = (uint8_t)sender_id;
+  tmpbuf[3] = (uint8_t)(sender_id >> 8);
+  tmpbuf[4] = (uint8_t)length;
+
+  u16 crc = crc16_ccitt(tmpbuf, sizeof(tmpbuf), 0);
+  crc = crc16_ccitt(buff, length, crc);
+  ck_assert(crc == 0xC13A);
```
yizow: can we put this crc constant as a static const array at the top of this function, just to pattern match the rest of the file.

c/tests/check_ubx.c
```diff
@@ -101,6 +101,28 @@ static void ubx_sbp_callback_rxm_rawx(
   msg_index++;
 }
 
+static void ubx_sbp_callback_rxm_sfrbx(
+    u16 msg_id, u8 length, u8 *buff, u16 sender_id, void *context) {
+  ck_assert(context);
+
+  int *cnt = (int *)context;
```
yizow: Is the point of `cnt` to make sure we only run the callback once for the test file?

If so, I think it'd be better if we either:
1. hex edit trimmed the test file so it contains _only one_ sfrbx message. It's currently a 69k file of which we're using a few hundred bytes?
2. Know exactly how many sfrbx messages we expect from the test file, do NOT return early if `*cnt > 0`, and then let the `ck_assert({expected_num_messages} == cnt)` catch it later. (I still think we can probably trim the test file a bit)

The idea here is that both the test file and the test should be extremely explicit in exactly what they expect. As is, `if (cnt > 0) {return;}` could potentially mask subsequent errors after the first.

c/include/gnss-converters/ubx_sbp.h
```diff
@@ -13,12 +13,13 @@
 #ifndef GNSS_CONVERTERS_UBX_SBP_INTERFACE_H
 #define GNSS_CONVERTERS_UBX_SBP_INTERFACE_H
 
-#include <unistd.h>
-
 #include <libsbp/gnss.h>
 #include <libsbp/navigation.h>
 #include <libsbp/observation.h>
 #include <libsbp/sbp.h>
+#include <unistd.h>
```
adel-mamin: `#include <unistd.h>` is moved by clang-format.

c/src/ubx_sbp.c
adel-mamin: The shuffling seems to be done by clang-format.

c/src/ubx_ephe/gps_l1ca.c
adel-mamin: Ok.

c/src/ubx_ephe/gps_l1ca.c
adel-mamin: I reworked it to keep the state in `ubx_sbp_state`.

c/src/ubx_ephe/gps_l1ca.c
adel-mamin: I removed `gps_l1ca_init()` altogether and use the callback from `ubx_sbp_state`.

c/src/CMakeLists.txt
adel-mamin: Done.

c/src/ubx_sbp.c
adel-mamin: Done. Please check if it matches your expectation.

c/src/ubx_ephe/ubx_ephe.c
adel-mamin: Done.

c/tests/check_ubx.c
adel-mamin: Done.

c/tests/check_ubx.c
adel-mamin: I trimmed the test file and simplified the test code. Please review it again.

c/src/ubx_sbp.c
yizow: Ah, got it. Huh, I wonder why it never changed it before.

PR #316, opened 2020-02-03
[SEN-85] Support IMU and ODO decoding from uBlox

reimerix:
This PR adds the capability do decode IMU data contained in `UBX-ESF-RAW` to the `IMU_RAW` SBP message ([SEN-85]), as well as decoding wheelspeed measurements contained in `UBX-ESF-MEAS` into the SBP `ODOMETRY` message. `ubx2sbp` will also periodically output the IMU configuration (`MSG_IMU_AUX`). `UBX-NAV-STATUS` will have to be enabled to produce valid GPS timestamps.

Besides this, formerly disabled test have been fixed and re-enabled.

[SEN-85]: https://swift-nav.atlassian.net/browse/SEN-85


c/include/gnss-converters/ubx_sbp.h
```diff
@@ -37,6 +37,15 @@ struct gps_sat_data {
   unsigned vmask;
 };
 
+/* Stores state for ESF-* messages */
+typedef struct ubx_esf_state {
```
jbangelo: It doesn't look like we're `typedef`ing other structs in this file, we should probably try and be consistent in style.

c/src/ubx_sbp.c
```diff
@@ -65,6 +67,8 @@
 #define SBP_OBS_DOPPLER_MASK (1 << 3)
 #define SBP_OBS_TRACK_STATE_MASK (0x07)
 
+static const double ubx_sensortime_scale = 39.0625e-6;
```
jbangelo: Is this from a particular model or are we thinking this is common in the UBX protocol? If it's model specific then a comment noting this would be helpful.

c/src/ubx_sbp.c
```diff
@@ -471,6 +494,290 @@ static int fill_msg_pos_llh_hnr(const u8 buf[], msg_pos_llh_t *msg) {
...
+  ubx_esf_meas esf_meas;
+  if (ubx_decode_esf_meas(inbuf, &esf_meas) != RC_OK) {
+    return;
+  }
+  msg_odometry_t msg_odo;
+  if (state->esf_state.tow_offset_set) {
+    double tow_s =
+        ubx_convert_msss_to_tow(esf_meas.calib_tag, &state->esf_state);
+    msg_odo.flags = 1;
+    msg_odo.tow = (u32)(tow_s * 1000.0);
```
jbangelo: Magic number, libswiftnav has a `SECS_MS` constant that can be used here.

c/src/ubx_sbp.c
```diff
@@ -471,6 +494,290 @@ static int fill_msg_pos_llh_hnr(const u8 buf[], msg_pos_llh_t *msg) {
...
+static void handle_esf_meas(struct ubx_sbp_state *state, u8 *inbuf) {
+  ubx_esf_meas esf_meas;
+  if (ubx_decode_esf_meas(inbuf, &esf_meas) != RC_OK) {
+    return;
+  }
+  msg_odometry_t msg_odo;
+  if (state->esf_state.tow_offset_set) {
+    double tow_s =
+        ubx_convert_msss_to_tow(esf_meas.calib_tag, &state->esf_state);
+    msg_odo.flags = 1;
```
jbangelo: Magic number, a named constant here would be good

c/src/ubx_sbp.c
```diff
@@ -471,6 +494,290 @@ static int fill_msg_pos_llh_hnr(const u8 buf[], msg_pos_llh_t *msg) {
...
+  int idx;
+  for (idx = 0; idx < num_meas; idx++) {
+    u32 data_type = ((u32)esf_meas.data[idx] & 0x3F000000) >> 24;
+    u32 data_value = ((u32)esf_meas.data[idx] & 0xFFFFFF);
+
+    switch (data_type) {
+      case ESF_REAR_LEFT_WHEEL_TICKS:
+        velocity_source = 1;
+        is_odo = true;
+        msg_odo.velocity = data_value & 0x3FFFFF;
```
jbangelo: It looks like we're masking and shifting `data_value` in the same way in each case, why not do it once up above?

c/src/ubx_sbp.c
```diff
@@ -471,6 +494,290 @@ static int fill_msg_pos_llh_hnr(const u8 buf[], msg_pos_llh_t *msg) {
...
+                           sizeof(msg_odo),
+                           (u8 *)&msg_odo,
+                           state->sender_id,
+                           state->context);
+    }
+  }
+}
+
+double ubx_convert_msss_to_tow(u32 msss, const ubx_esf_state_t *state) {
+  double tow = 0.001 * msss + state->time_since_startup_tow_offset;
```
jbangelo: Magic number, could be `msss / (double)SECS_MS`

c/src/ubx_sbp.c
```diff
@@ -471,6 +494,290 @@ static int fill_msg_pos_llh_hnr(const u8 buf[], msg_pos_llh_t *msg) {
...
+  if (!state->tow_offset_set) {
+    return -1;
+  } else if (msss < state->last_sync_msss - 100) {
+    /* Rollover occured since last sync */
+    const u32 msss_max = 0xFFFFFFFF;
```
jbangelo: Magic number, should probably use `UINT32_MAX`

c/src/ubx_sbp.c
```diff
@@ -471,6 +494,290 @@ static int fill_msg_pos_llh_hnr(const u8 buf[], msg_pos_llh_t *msg) {
...
+    const double time_since_startup_rollover = 0.001 * msss_max + 0.001;
+    tow += time_since_startup_rollover;
+    return tow;
```
jbangelo: Duplicated `return` here, if you're just going to return immediately after this else block there's no need to return inside of it.

c/src/ubx_sbp.c
```diff
@@ -471,6 +494,290 @@ static int fill_msg_pos_llh_hnr(const u8 buf[], msg_pos_llh_t *msg) {
...
+static void maybe_parse_imu_data(u32 data,
+                                 u8 data_type,
+                                 msg_imu_raw_t *msg,
+                                 ubx_esf_state_t *esf_state) {
+  s32 parsed_value = convert24b(data);
+  if (data_type == ESF_X_AXIS_GYRO_ANG_RATE ||
+      data_type == ESF_Y_AXIS_GYRO_ANG_RATE ||
+      data_type == ESF_Z_AXIS_GYRO_ANG_RATE) {
+    const float angular_rate_degs = ldexpf(1, -12) * parsed_value;
+    const float scale = (float)(125.0 / 32768.0);
```
jbangelo: Magic numbers again

c/src/ubx_sbp.c
```diff
@@ -471,6 +494,290 @@ static int fill_msg_pos_llh_hnr(const u8 buf[], msg_pos_llh_t *msg) {
...
+  msg->tow = timespec.tow;
+  msg->tow_f = timespec.tow_f;
+
+  if (!esf_state->tow_offset_set) {
+    const u32 time_invalid_bit = (1 << 31);
+    msg->tow |= time_invalid_bit;
+  }
+}
+
+static bool check_imu_message_complete(const s16 *received_number_msgs,
```
jbangelo: It looks like if we ever drop one message out of the 6 IMU messages we'll forever be out of sync. Is that true?

c/src/ubx_sbp.c
```diff
@@ -519,6 +826,37 @@ static void handle_nav_velecef(struct ubx_sbp_state *state, u8 *inbuf) {
...
+                   (nav_status.status_flags & tow_ok);
+
+  if (gnss_fix_good && time_good) {
+    state->esf_state.time_since_startup_tow_offset =
+        0.001 * nav_status.i_tow - 0.001 * nav_status.msss;
+    state->esf_state.tow_offset_set = true;
+  }
+}
+
+static void handle_nav_clock(u8 *inbuf) {
```
jbangelo: Why do we need this function? It appears to be doing nothing.

reimerix: Right, that slipped through from when we were still trying different things. Will delete it.

c/src/ubx_sbp.c
reimerix: To answer your question: We'll only be out of sync for this burst of IMU data. Everything will start anew with the next `UBX-ESF-RAW` message.

A burst of IMU data is contained in the `UBX-ESF-RAW` message. It's usually 9-11 times the 6-axis data of consecutive instants of time. With all the datasets we have at hand, we have never observed that there would be incomplete IMU data for one timestamp before moving on to the next 6-axis packet in this` UBX-ESF-RAW`.

Given that we have never observed incomplete IMU data with the datasets we have, I'd be in favour of not handling that case. I'll make that explicit by throwing in a couple of asserts - let me know if you think a different method would be better.

jbangelo: It's difficult if the UBX protocol doesn't provide a good way to detect when some of the IMU messages are dropped. It shouldn't be a problem if the UBX data is sent over a reliable stream (e.g. TCP), but if it's coming over an unreliable stream (e.g. serial) it's definitely possible to drop messages.

If there isn't much we can do to reliably detect it at the protocol level in real time this should probably go on the list of things we'd want to be able to detect with a post processing diagnostic tool.

jbangelo: I guess this would be covered by a more general check to see if there were any missing IMU messages, so no need for a separate tool or check.

reimerix: +1 for the fact that at some point, there probably will be a problem with either a complete `UBX-ESF-RAW` going lost due to e.g. bit errors or IMU packets inside this message going missing because the uBlox receiver couldn't receive or decode them. While the risk for the latter is hard to quantify without knowing implementation details, I wouldn't think the probability would be too high because the receiver and the IMU are connected via SPI, and are presumably built into one shielded module.

As for diagnostic tools: a tool checking for large gaps in the IMU stream is on our list, and rare little gaps shouldn't impose a big problem.

Thanks for all this really valuable feedback, @jbangelo !

PR #317, opened 2020-02-04
Add option to force output COG field in VTG sentence

pmiettinen:


pmiettinen:
`codecov/project/gnss-converters — 81.16% (-0.01%) compared to 14b4683` is due to added value assignment in an if branch that wasn't tested before either. To cover it would require reworking the test data.

c/src/nmea.c
```diff
@@ -69,6 +69,11 @@
 #define NMEA_COG_STATIC_LIMIT_MS 0.1f
 #define NMEA_COG_STATIC_LIMIT_KNOTS MS2KNOTS(NMEA_COG_STATIC_LIMIT_MS, 0, 0)
 #define NMEA_COG_STATIC_LIMIT_KPH MS2KMHR(NMEA_COG_STATIC_LIMIT_MS, 0, 0)
+#define NMEA_COG_FORCED_INIT 0
+
+static bool vtg_cog_forced = false;
```
nsirola: This should rather go into `state` to allow for parallel instances

c/src/nmea.c
```diff
@@ -734,6 +739,8 @@ void send_gpvtg(const sbp2nmea_t *state) {
   const msg_vel_ned_t *sbp_vel_ned =
       sbp2nmea_msg_get(state, SBP2NMEA_SBP_VEL_NED);
 
+  static double cog_pinned = NMEA_COG_FORCED_INIT;
```
nsirola: Same here

c/src/nmea.c
pmiettinen: Makes sense.

c/src/nmea.c
pmiettinen: Problem with `state` being qualified as `const`. I'll try to figure out something.

pmiettinen: No easy and correct solution readily in mind. Couple of possible but questionable approaches.

1. Gun-to-the-head approach, just cast to non-const locally inside the function in question.
2. Remove `const` from the the `send` function signature. Would affect all sentences' send functions.
3. Decoupling calculation and sending would lead to unreasonable amount of code rework considering the dubiousness of the added feature in the first place.

nsirola: or, add `cog` and `sog` to the state struct `sbp2nmea_t` and compute them already somewhere in the velocity message callback?

c/src/sbp_nmea.c
```diff
@@ -158,7 +158,7 @@ static void check_nmea_send(sbp2nmea_t *state) {
       continue;
     }
 
-    nmea_meta[id].send(state);
+    nmea_meta[id].send(state, &state->pinned);
```
pmiettinen: Part of data is referenced twice but at least it is clear which part might get modified.

`sbp2nmea_t` could be restructured to enable following of JPL rule#6 "Restrict the scope of data to the smallest possible.".

c/src/nmea.c
```diff
@@ -310,7 +310,7 @@ void get_utc_time_string(bool time,
  *
  * \param state Current SBP2NMEA state
  */
-void send_gpgga(const sbp2nmea_t *state) {
+void send_gpgga(const sbp2nmea_t *state, sbp2nmea_pinned_t *pinned) {
```
pmiettinen: Need to update the doc strings.

c/src/nmea.c
pmiettinen: By "velocity message callback" do you mean `vel_ned_callback` in `sbp_nmea_bridge`? That would be actually a step in the right direction but as a one-of change would make an oddball comparing to other sentences/callback.

Currently the SBP and NMEA are too tightly coupled. More suitable would be to have something like:
```
 SBP
    \
     ---> COG/SOG etc --> NMEA
    /
RTCM
Whatever input format
```

c/include/gnss-converters/nmea.h
```diff
@@ -44,15 +44,15 @@
...
-void send_gpgga(const sbp2nmea_t *state);
-void send_gprmc(const sbp2nmea_t *state);
-void send_gpvtg(const sbp2nmea_t *state);
-void send_gpgll(const sbp2nmea_t *state);
-void send_gpzda(const sbp2nmea_t *state);
-void send_gpgst(const sbp2nmea_t *state);
-void send_gphdt(const sbp2nmea_t *state);
-void send_gsa(const sbp2nmea_t *state);
-void send_gsv(const sbp2nmea_t *state);
+void send_gpgga(const sbp2nmea_t *state, sbp2nmea_pinned_t *pinned);
```
anth-cole: I don't see why we need to change the function parameter - couldn't `sbp2nmea_pinned_t *pinned` which is just a double, just become a field in the sbp2nmea_t struct? this would significantly reduce the footprint of the change I think

c/src/sbp_nmea.c
```diff
@@ -274,7 +274,13 @@ void sbp2nmea_soln_freq_set(sbp2nmea_t *state, float soln_freq) {
...
 }
 
+void sbp2nmea_vtg_cog_forced_set(sbp2nmea_t *state, bool forced) {
+  state->vtg_cog_forced = forced;
+}
+
 void sbp2nmea_init(sbp2nmea_t *state, void (*cb_sbp_to_nmea)(u8 msg_id[])) {
   memset(state, 0, sizeof(*state));
+  state->vtg_cog_forced = false;
+  state->pinned.vtg_cog_deg = NMEA_COG_FORCED_INIT;
```
anth-cole: I'm very nervous about this initialization to 0 - I agree with you @pmiettinen that this should be sent as an empty field before our first valid cog is received.

c/include/gnss-converters/nmea.h
pmiettinen: Did you check this related conversation https://github.com/swift-nav/gnss-converters/pull/317#discussion_r374588961

anth-cole: oops - sorry I missed that conversation. imo I'd go with option 2 as the least bad option - we have the same pattern across other converters that the 'state' is non-const, I think that creates the least problems here imo

pmiettinen: ok, I'll do the change

pmiettinen:
Updated to latest design after yesterday's discussion.

c/include/gnss-converters/sbp_nmea.h
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018 Swift Navigation Inc.
+ * Copyright (C) 2018, 2020 Swift Navigation Inc.
```
anth-cole: nit: can we not make this change? we haven't been doing it elsewhere and general policy is to leave the date as the date of file creation.)

c/include/gnss-converters/nmea.h
```diff
@@ -1,6 +1,6 @@
 /*
- * Copyright (C) 2013-2017 Swift Navigation Inc.
- * Contact: Fergus Noble <fergus@swift-nav.com>
+ * Copyright (C) 2013-2017, 2020 Swift Navigation Inc.
```
anth-cole: Let's change the email contact, but leave the date as is

c/src/nmea.c
```diff
@@ -1,6 +1,6 @@
 /*
- * Copyright (C) 2013-2017 Swift Navigation Inc.
- * Contact: Fergus Noble <fergus@swift-nav.com>
+ * Copyright (C) 2013-2017, 2020 Swift Navigation Inc.
```
anth-cole: same comment as above

c/src/nmea.c
```diff
@@ -308,7 +302,7 @@ void get_utc_time_string(bool time,
 /** Assemble a NMEA GPGGA message and send it out NMEA USARTs.
  * NMEA GPGGA message contains Global Positioning System Fix Data.
  *
- * \param state Current SBP2NMEA state
+ * \param state  Current SBP2NMEA state
```
anth-cole: nit: extra space

c/src/sbp_nmea.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2018-2019 Swift Navigation Inc.
+ * Copyright (C) 2018-2020 Swift Navigation Inc.
```
anth-cole: nit: same as above

c/tests/check_nmea.c
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 Swift Navigation Inc.
+ * Copyright (C) 2017, 2020 Swift Navigation Inc.
```
anth-cole: nit: same as above

c/tests/nmea_truth.h
```diff
@@ -1,6 +1,6 @@
 /*
- * Copyright (C) 2016 Swift Navigation Inc.
- * Contact: Pasi Miettinen <pasi.miettinen@exafore.com>
+ * Copyright (C) 2016, 2020 Swift Navigation Inc.
```
anth-cole: nit: same here

c/src/nmea.c
```diff
@@ -702,7 +693,7 @@ void send_gprmc(const sbp2nmea_t *state) {
 
   if ((sbp_vel_ned->flags & VELOCITY_MODE_MASK) != VELOCITY_MODE_NONE) {
     NMEA_SENTENCE_PRINTF("%.2f,", sog_knots); /* Speed */
-    if (NMEA_COG_STATIC_LIMIT_KNOTS < sog_knots) {
+    if (MS2KNOTS(state->cog_threshold_ms, 0, 0) <= sog_knots) {
```
anth-cole: I can't see where `MS2KNOTS` is defined?

c/include/gnss-converters/sbp_nmea.h
pmiettinen: Sure thing.

c/src/nmea.c
pmiettinen: Macro is defined in `nmea.h`.

c/include/gnss-converters/sbp_nmea.h
```diff
@@ -78,6 +78,8 @@ typedef struct sbp2nmea_state {
 
   float soln_freq;
 
+  float cog_threshold_ms;
```
switanis: ms is quite confusing, I immediately read this as milliseconds.
mps would be better

c/src/nmea.c
```diff
@@ -702,7 +693,7 @@ void send_gprmc(const sbp2nmea_t *state) {
 
   if ((sbp_vel_ned->flags & VELOCITY_MODE_MASK) != VELOCITY_MODE_NONE) {
```
switanis: When we do not have velocity? That can still affect NMEA output even when the COG output speed threshold is set to zero (meaning always output COG).

c/src/nmea.c
switanis: That macro:
#define MS2KNOTS(x, y, z) (sqrt((x) * (x) + (y) * (y) + (z) * (z)) * 1.94385)

is very inefficient, making unnecessary heavy float operations.
Suggesting to make a new simple macro.

c/include/gnss-converters/sbp_nmea.h
pmiettinen: True, mps is more suitable. I'll change that.

c/src/nmea.c
pmiettinen: Probably when there's no fix. I'll try to see in more detail tomorrow.

c/src/nmea.c
pmiettinen: Sure, I'll add an alternative macro for one dimensional vector.

Although.. `MS2KNOTS` is called from RMC and VTG sentence functions twice from both. Default rates being 10Hz for RMC and 1Hz for VTG. Totalling 22 calls per second. I think it's like worrying that one more bucket will flood the ocean. Also for GCC optimizing `0 * 0` away should be a cakewalk.

c/src/nmea.c
pmiettinen: Assumption confirmed here https://github.com/swift-nav/starling/blob/master/src/pvt_engine/firmware_binding.cc#L128

```C
  // The filter in its current state (July 2018) always uses computed
  // Doppler to observe velocity. We hard code this in for now.
  result->flags.velocity_mode = VELOCITY_MODE_COMPUTED_DOPPLER;
```
So this would mean that the exact description for the COG threshold value `0` with this implementation would be along the lines of "Output NMEA COG fields always during valid fix".

switanis: ok

switanis:
Will try today.

PR #319, opened 2020-02-07
Added BDS D1 UBX ephemeris decoder

adel-mamin:
The planning issue https://swift-nav.atlassian.net/browse/STAR-953
The PR adds support for ephemeris conversion from UBX to SBP format.
The ephemeris data is extracted from Beidou D1 navigation message.

The PR is contingent on https://github.com/swift-nav/libswiftnav/pull/119


pgrgich:
I think there's a fair bit of overlap with https://github.com/swift-nav/gnss-converters/pull/321 here, since uBlox seems to just pass through the original nav message with a small wrapper around it, is that right?

adel-mamin:
> I think there's a fair bit of overlap with #321 here, since uBlox seems to just pass through the original nav message with a small wrapper around it, is that right?

Yes, uBlox nav message is in the broadcast navigation data as described in https://www.u-blox.com/sites/default/files/products/documents/u-blox8-M8_ReceiverDescrProtSpec_(UBX-13003221)_Public.pdf on the page 29. The wrapper says the GNSS id and satellite id (page 368, same document).
https://github.com/swift-nav/gnss-converters/pull/321 seems to convert from SBP to RTCM, whereas this PR converts from UBX to SBP.

PR #320, opened 2020-02-11
standardize write callback function signature

yizow:
Needed to enable IO Swiftlet modularity.

Changes return from `void` to `int`, which shouldnt affect any functionality.


c/include/gnss-converters/sbp_rtcm3.h
```diff
@@ -42,7 +42,7 @@ struct rtcm3_out_state {
   s8 leap_seconds;
   bool leap_second_known;
   bool ant_known;
-  void (*cb_sbp_to_rtcm)(u8 *buffer, u16 length, void *context);
+  int (*cb_sbp_to_rtcm)(u8 *buffer, u16 length, void *context);
```
woodfell: Could you use well defined types rather than `int` please, `int` has a platform specific size. I think `s32` or `int32_t` would be in keeping with `libsbp`

yizow: Fixed

PR #321, opened 2020-02-13
Implement GLO and BDS ephemeris conversion from SBP to RTCM.

pgrgich:
Looking to get some eyes on it to try to speed the merge for Tesla support.

GLONASS ICD: 
[ICD_GLONASS_4.0_(1998)_en.pdf](https://github.com/swift-nav/gnss-converters/files/4200135/ICD_GLONASS_4.0_.1998._en.pdf)

BEIDOU ICD:
[beidou_icd_english_ver2.0.pdf](https://github.com/swift-nav/gnss-converters/files/4200181/beidou_icd_english_ver2.0.pdf)

RTCM ICD:
[RTCM 3.3 - c10403.3 - DGNSS.PDF](https://github.com/swift-nav/gnss-converters/files/4200137/RTCM.3.3.-.c10403.3.-.DGNSS.PDF)

SBP message structure: https://github.com/swift-nav/libsbp/blob/master/c/include/libsbp/observation.h#L574

RTCM message structure: https://github.com/swift-nav/librtcm/blob/master/c/include/rtcm3/messages.h#L263

CC @reimerix, feel free to look it over, since you love RTCM so much. :)


c/src/sbp_rtcm3_ephemeris.c
```diff
@@ -70,6 +70,97 @@ void sbp_to_rtcm3_gps_eph(const msg_ephemeris_gps_t *sbp_gps_eph,
...
+                          const struct rtcm3_out_state *state) {
+  (void)state;
+  assert(sbp_glo_eph);
+  assert(msg_eph);
+  assert(is_glo(sbp_glo_eph->common.sid.code));
+
+  msg_eph->sat_id = sbp_gps_eph->common.sid.sat;
+  msg_eph->constellation = CONSTELLATION_GLO;
+  msg_eph->wn = sbp_glo_eph->common.toe.wn % 1024;
+  msg_eph->toe = sbp_glo_eph->common.toe.tow / GPS_TOE_RESOLUTION;
```
pgrgich: This is inevitably wrong. I'm hopeful someone can provide some information while I move on with other implementation. :)

jbangelo: Looking at the decoding data, it looks like the GLONASS RTCM message doesn't contain a TOE field? If that's the case the value in this field probably isn't needed for encoding a GLONASS RTCM message. Putting something here is probably the right thing to do, and it appears the decoding code calculates the TOE in GPS time based on the latest time seen in an observation message. I think this might be good enough? Not entirely sure about that though.

pgrgich: The RTCM message a bunch of time parameters that are defined by the GLO ICD (t_k, t_b and N_T, at least). I can try to piece it all together, but I was hoping someone might know what to do. :)

c/src/rtcm3_utils.c
```diff
@@ -743,6 +743,95 @@ uint8_t convert_uri_to_ura(float uri) {
...
+    ura = 14;
+  } else if (fabs(uri - 6144.0f) < FLOAT_EQUALITY_EPS) {
+    ura = 15;
+  } else {
+    ura = (uint8_t)(log2f(uri) + 2.0);
+  }
+  return ura;
+}
+
+float convert_glo_ft_to_meters(const uint8_t ft) {
```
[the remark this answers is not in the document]
pgrgich: Nope, that was a mistake.

c/src/sbp_rtcm3.c
```diff
@@ -1024,6 +1024,42 @@ void sbp2rtcm_sbp_gps_eph_cb(const u16 sender_id,
...
+  (void)len;
+  (void)sender_id;
+  msg_ephemeris_glo_t *sbp_glo_eph = (msg_ephemeris_glo_t *)msg;
+
+  rtcm_msg_eph msg_glo_eph;
+  u8 frame[RTCM3_MAX_MSG_LEN];  // Max RTCM message length is 1023 Bytes
+
+  /* generate and send the glo ephemeris message */
+  sbp_to_rtcm3_glo_eph(sbp_glo_eph, &msg_glo_eph, state);
+  u16 frame_size = encode_rtcm3_frame(&msg_glo_eph, 1020, frame);
```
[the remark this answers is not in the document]
pgrgich: Aha! That was the bit I was missing. :)

c/src/sbp_rtcm3_ephemeris.c
jbangelo: Yeah, I can't help you there other than pointing to the existing code that does the RTCM->SBP conversion which does the inverse operation.

ljbade: GLONASS likes to be different so it uses different names and idea then everyone else.

N_T with t_k is equivalent to toe, except the encoding breaks things down into days, hours, minutes and seconds since the last leap year so you need to carefully read how they are defined in the ICD

ljbade:
One thing to note is that you will probably find going from SBP to RTCM won't be lossless since there are a number of fields in the native ephemerides that Piksi had no use for so we just didn't bother putting them in the SBP message

c/src/sbp_rtcm3_ephemeris.c
pgrgich: I think I've got it so that we will encode what is necessary for us to decode and work properly, but not convinced that it will suffice for others. I'd appreciate you guys checking it out again.

c/src/sbp_rtcm3_ephemeris.c
```diff
@@ -70,6 +71,144 @@ void sbp_to_rtcm3_gps_eph(const msg_ephemeris_gps_t *sbp_gps_eph,
...
+  msg_eph->wn = sbp_glo_eph->common.toe.wn;
+  msg_eph->toe = sbp_glo_eph->common.toe.tow;
+  msg_eph->ura = convert_glo_uri_to_ura(sbp_glo_eph->common.ura);
+
+  msg_eph->fit_interval =
+      rtcm3_encode_fit_interval_glo(sbp_glo_eph->common.fit_interval);
+  msg_eph->health_bits = sbp_glo_eph->common.health_bits;
+  gps_time_t time = sbp_gps_time_2_gps_time(&sbp_glo_eph->common.toe);
+  if (!compute_glo_time_of_day(&time, state, &msg_eph->glo.t_b)) {
+    msg_eph->health_bits = 1;
```
jbangelo: Do we want to completely clear out the previous value? Without knowing the meaning of `health_bits` I would assume we would want to set a specific bit to 1 leaving the others as they were previously set.

pgrgich: `health_bits` is a single bit with the very intuitive definition of 0 is healthy and non-zero is unhealthy. I'm working on the basis that any problem causes us not to be healthy.

PR #322, opened 2020-02-18
[SEN-137] Adjust M8L IMU timestamp

reimerix:
This PR adjusts the M8L IMU timestamp by the 50 ms offset which has been found empirically by comparing the M8L IMU angular rate  with a properly timestamped reference IMU.

### DIFF
--- a/c/src/ubx_sbp.c
+++ b/c/src/ubx_sbp.c
@@ -889,6 +889,7 @@ static void handle_nav_status(struct ubx_sbp_state *state, u8 *inbuf) {
   if (gnss_fix_good && time_good) {
     state->esf_state.time_since_startup_tow_offset =
         0.001 * nav_status.i_tow - 0.001 * nav_status.msss;
+    state->esf_state.last_sync_msss = nav_status.msss;
     state->esf_state.tow_offset_set = true;
   }
 }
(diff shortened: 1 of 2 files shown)


c/src/ubx_sbp.c
```diff
@@ -876,6 +887,7 @@ static void handle_nav_status(struct ubx_sbp_state *state, u8 *inbuf) {
   if (gnss_fix_good && time_good) {
     state->esf_state.time_since_startup_tow_offset =
         0.001 * nav_status.i_tow - 0.001 * nav_status.msss;
+    state->esf_state.last_sync_msss = nav_status.msss;
```
richarddeurloo: I guess this fixes a possible undetected roll-over ?

reimerix: It does, and it was just missing before. The logic for handling those rollovers is tested separately in https://github.com/swift-nav/gnss-converters/blob/7a4e72c14c78d2c01f71f534ee8ea372ecd54e34/c/tests/check_ubx.c#L535 .

c/src/ubx_sbp.c
```diff
@@ -710,6 +710,17 @@ static void set_sbp_imu_time(u32 sensortime_this_message,
     sensortime = sensor_tow_s;
   }
 
+  // This constant has been found empirically by comparing the M8L IMU angular
+  // rate with a properly time stamped reference.
+  const double ubx_gnss_time_error = 0.05;
```
jbangelo: Nit: unless we know this offset is unintentional this value should probably be called an offset instead of an error.

PR #330, opened 2020-02-27
Added GAL I/NAV UBX ephemeris decoder

adel-mamin:
The planning issue https://swift-nav.atlassian.net/browse/STAR-953
The PR adds support for ephemeris conversion from UBX to SBP format.
The ephemeris data is extracted from Galileo I/NAV navigation message.

The PR is contingent on https://github.com/swift-nav/libswiftnav/pull/124


c/src/ubx_ephemeris/gal.c
```diff
@@ -0,0 +1,168 @@
...
+  msg->inc = k->inc;
+  msg->inc_dot = k->inc_dot;
+  msg->af0 = k->af0;
+  msg->af1 = (float)k->af1;
+  msg->af2 = (float)k->af2;
+  msg->toc.tow = (u32)round(k->toc.tow);
+  msg->toc.wn = k->toc.wn;
+  msg->iode = k->iode;
+  msg->iodc = k->iodc;
+  /* I/NAV message as per UBX-13003221-R17
```
nsirola: I'd take the setting of source outside this function, so this remains a generic helper (and candidate for later refactoring) without UBX dependencies

adel-mamin: Agree. Fixed.

c/src/ubx_ephemeris/gal.c
```diff
@@ -0,0 +1,181 @@
...
+  sat->vmask |= 1U << (wtype - 1);
+  assert(wtype <= (int)ARRAY_SIZE(sat->pg));
+  memcpy(&sat->pg[wtype - 1].words, words, sizeof(sat->pg[wtype - 1].words));
+  if (ALL_PAGES_MASK != (sat->vmask & ALL_PAGES_MASK)) {
+    return; /* not all word types 1,2,3,4,5 are available yet */
+  }
+
+  int iod[4];
+  for (int i = 0; i < (int)ARRAY_SIZE(iod); i++) {
+    iod[i] = ((words[0] >> 16U) << 2) & 0x3FF;
```
jbangelo: This seems like an indirect way of clearing out the lower two bits. Why not just do
```suggestion
    iod[i] = (words[0] >> 14U) & 0x3FC;
```

c/src/ubx_ephemeris/gal.c
```diff
@@ -0,0 +1,181 @@
...
+    iod[i] |= ((words[0] >> 14U) & 3);
```
jbangelo: Nit: we should be consistent in how we represent mask values. This should probably be `0x3` instead.

c/src/common.h
```diff
@@ -19,6 +19,12 @@
 #include "swiftnav/ephemeris.h"
 #include "swiftnav/signal.h"
 
+#if defined(__clang__) || defined(__GNUC__) || defined(__GNUG__)
+#define ASSERT_STATIC(expr) _Static_assert(expr, "")
+#else
+#error "Define ASSERT_STATIC()"
+#endif
```
jbangelo: We should avoid any patterns that prevent this code from being compiled with standards conforming compilers. Do we really need a static assert over a regular runtime assert?

c/include/gnss-converters/ubx_sbp.h
```diff
@@ -31,9 +31,14 @@ extern "C" {
 #define DEFAULT_UBX_SENDER_ID 61568
 
 struct sat_data {
-  struct subframe {
-    u32 words[10];
-  } sf[3];
+  union {
```
jbangelo: Is it really helpful to use the same `sat_data` struct for GAL sats? It seems like there isn't ever a place where we would need to reuse a GAL `sat_data` as a GPS or BDS `sat_data`. Plus this union unnecessarily inflates the GPS and BDS structs by 40 bytes.

A dedicated `gal_sat_data` I think would be preferable here.

c/src/ubx_ephemeris/gal.c
adel-mamin: Sure. My bad. I will fix it.

c/include/gnss-converters/ubx_sbp.h
adel-mamin: Agree. I will fix it.

c/src/common.h
adel-mamin: No an absolute need for this. I will replace it with a run-time assert.

c/src/ubx_ephemeris/gal.c
adel-mamin: The lines 124 and 125 are now merged into one line.

c/src/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ add_library(
   ubx_sbp.c
   ${PROJECT_SOURCE_DIR}/src/ubx_ephemeris/gps.c
   ${PROJECT_SOURCE_DIR}/src/ubx_ephemeris/bds.c
+  ${PROJECT_SOURCE_DIR}/src/ubx_ephemeris/gal.c
+  ${PROJECT_SOURCE_DIR}/third_party/libswiftnav/src/ephemeris.c
```
nsirola: This shouldn't be needed if you rebase on top of master (see https://github.com/swift-nav/gnss-converters/pull/331)

c/src/ubx_ephemeris/gal.c
```diff
@@ -0,0 +1,180 @@
...
+    iod[i] = (words[0] >> 14U) & 0x3FFU;
+  }
+
+  for (int i = 0; i < (int)ARRAY_SIZE(iod); i++) {
+    for (int j = i + 1; j < (int)ARRAY_SIZE(iod); j++) {
+      if (iod[i] != iod[j]) {
+        return; /* received pages are not from the same batch */
+      }
+    }
+  }
```
nsirola:
```suggestion
  for (int i = 1; i < (int)ARRAY_SIZE(iod); i++) {
      if (iod[i] != iod[0]) {
        return; /* received pages are not from the same batch */
      }
  }
```

nsirola: Should be enough to compare that all are equal to the first one?

c/src/CMakeLists.txt
adel-mamin: True. I will fix it.

c/src/ubx_ephemeris/gal.c
adel-mamin: Agree. I will fix it.

PR #331, opened 2020-03-01
Remove third party paths

woodfell:
There are a few hard coded paths in the gnss-converters makefile which are polluting upstream include paths. A file from libswiftnav `src/ephemeris.c` is actually being built directly in to gnss-converters rather than just linking against the library.

Confusingly the version of libswiftnav checked out as a submodule in this project was never being included because the version from libubx was being found first. This is a danger when using submodules and the first come first served dependency resolution system required by cmake. Getting auto submodule update enabled for libswiftnav would probably help prevent these things, for the moment I have moved a few command in CMakeLists.txt to prevent the error.


c/src/CMakeLists.txt
```diff
@@ -16,7 +16,6 @@ add_library(
   ubx_sbp.c
   ${PROJECT_SOURCE_DIR}/src/ubx_ephemeris/gps.c
   ${PROJECT_SOURCE_DIR}/src/ubx_ephemeris/bds.c
-  ${PROJECT_SOURCE_DIR}/third_party/libswiftnav/src/ephemeris.c
```
anth-cole: This seems out of line with good practice - any reason this might have been some necessary workaround or are you confident it was just a mistake?

woodfell: Probably just a mistake. Submodules were the downfall here because the version of libswiftnav included in this repo wasn't being built - updating the submodule here would do absolutely nothing. The version of libswiftnav that is a submodule in libubx was the one being built. If we get autosubmodule update working for libswiftnav it would probably help prevent this

PR #345, opened 2020-03-13
ESD-1757: introduced nmea callback function with context object as we…

RReichert:
PR introduces two elements, 1 bug fix and 1 new feature.

The bug fix is quite simple, when trying to call any of the function within `nmea.h` from a C++ code base, it fails to link because the C++ reads that header files as if it follows the C++ style function name mangling, which isn't the case as the source file is written in C.

The feature addition was again quite simple, originally NMEA had a callback function which wasn't taking any arguments. This is a bit limiting because that would mean that whatever is performed within the callback needs to call static variables, which with regards to our swiftlets work isn't good enough as we might have multiple sbp2nmea states.

**NOTE**: if there anyone else that I need to add to this PR, please let me know, I just added in what github has recommended to me

### DIFF
--- a/c/include/gnss-converters/sbp_nmea.h
+++ b/c/include/gnss-converters/sbp_nmea.h
@@ -79,6 +79,8 @@ typedef struct sbp2nmea_state {
   float soln_freq;
 
   void (*cb_sbp_to_nmea)();
+  void (*cb_sbp_to_nmea_ctx)(char *msg, void* ctx);
+  void* ctx;
 } sbp2nmea_t;
 
 #ifdef __cplusplus
@@ -86,6 +88,8 @@ extern "C" {
 #endif
 
 void sbp2nmea_init(sbp2nmea_t *state, void (*cb_sbp_to_nmea)(u8 msg_id[]));
+
+void sbp2nmea_ctx_init(sbp2nmea_t *state, void (*cb_sbp_to_nmea_ctx)(char* msg, void* ctx), void* ctx);
 
 void sbp2nmea(sbp2nmea_t *state,
               u8 len,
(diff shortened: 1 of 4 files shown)


c/include/gnss-converters/sbp_nmea.h
```diff
@@ -79,6 +79,8 @@ typedef struct sbp2nmea_state {
   float soln_freq;
 
   void (*cb_sbp_to_nmea)();
+  void (*cb_sbp_to_nmea_ctx)(char *msg, void *ctx);
```
nsirola: I think I'd rather add the context parameter to the existing callback and just pass nullptr wherever the context is not needed, instead of adding two parallel callbacks. Granted, that would change the API and require a small change everywhere this is used (only piksi_apps sbp_nmea_bridge so far I think), but the end result would be neater.

c/tests/check_nmea.c
```diff
@@ -45,74 +45,84 @@ static bool nmea_gsa_processed = false;
 static bool nmea_gpgst_processed = false;
 static bool nmea_gpgsv_processed = false;
 
+enum nmea_callback_index {
```
nsirola: Why not use the existing enum https://github.com/swift-nav/gnss-converters/blob/5d75a55cdffb3c99d4d569a710b28a9979cca21a/c/include/gnss-converters/sbp_nmea.h#L25

RReichert: Good call, didn't realize there was an enum value already for this. Made the changes.

c/include/gnss-converters/sbp_nmea.h
RReichert: agree with you on that, I just thought there would have been more locations where this might have been used, but since its only one, I'll break API and fix it in the auto submodule updates, should make things a lot cleaner.

RReichert: agree with you on that, I just thought there would have been more locations where this might have been used, but since its only one, I'll break API and fix it in the auto submodule updates, should make things a lot cleaner.

PR #348, opened 2020-03-18
[STAR-953] Add ubx->sbp SBAS raw data conversion

nsirola:
Convert the UBX-RXM-SFRBX  SBAS subframe message into SBP_MSG_SBAS_RAW.

In order to fill the timestamp for the message, had to add TOW field to the UBX state.

## Testing
Converted a short ubx log into SBP and run it through Starling `run-filter` without SBAS decoding related errors. SBAS corrections are being added into the SBAS manager, but was not able to check they are used into position solution correctly because the measurements in the log did not produce solutions at all.


c/src/ubx_ephemeris/sbas.c
```diff
@@ -0,0 +1,95 @@
...
+  assert(SBAS_PREAMBLE == preamble);
+
+  u8 message_type = (words[0] >> 18U) & 0x3FU;
+
+  u32 tow_ms = data->last_tow_ms;
+  /* the time stamp of the SBAS message is approximately the current time
+   * minus 1 s (message length) and an average of 120ms transit time */
+  tow_ms -= 1120;
+  if (tow_ms < 0.0) {
+    tow_ms += WEEK_MS;
```
adel-mamin: `tow_ms` is not used anywhere?

nsirola: True, fixed the below call to `pack_sbas_data` to use it as was intended.
Interesting that compiler warnings about that are not on for this project.

c/src/ubx_ephemeris/sbas.c
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  u32 tow_ms = data->last_tow_ms;
+  if (tow_ms > WEEK_MS) {
+    /* TOW not yet set */
+    return;
+  }
+
+  /* the time stamp of the SBAS message is approximately the current time
+   * minus 1 s (message length) and an average of 120ms transit time */
```
jbangelo: How were these values (1 second, 120 ms) arrived at? If they are from a UBlox data sheet I'd be inclined to trust them more than if they were determined by experiment.

Not to say that I don't trust these values, but it'd be good to note their original source so if something stops working here later we can a bit more context for these magic values.

c/src/ubx_ephemeris/sbas.c
jbangelo: Looks like `tow_ms` is still not used?

c/tests/check_ubx.c
```diff
@@ -562,6 +573,15 @@ START_TEST(test_rxm_sfrbx_gal) {
 }
 END_TEST
 
+START_TEST(test_rxm_sfrbx_sbas) {
+  struct ubx_sbp_state state;
+  ubx_sbp_init(&state, ubx_sbp_callback_rxm_sfrbx_sbas, NULL);
+  state.last_tow_ms = 250709100;
```
jbangelo: It would be good to add checks of this value for the unit tests for the messages that should be setting it.

c/src/ubx_ephemeris/sbas.c
nsirola: Augh, seems my workspaces are bit out of sync. Fixed again.

c/src/ubx_ephemeris/sbas.c
nsirola: Reworked the comment

c/tests/check_ubx.c
nsirola: Good point, added checks after the nav messages that this field is set.

c/src/ubx_ephemeris/sbas.c
nsirola: I'll add a PR to fix clang-tidy to catch these next.

benjaminaltieri:
this seems to have causes and error in `piksi_apps` builds - @nsirola can you take a look? https://github.com/swift-nav/piksi_apps/pull/530

nsirola:
Fixed and bumped `piksi_apps` `gnss-converters` pointer in https://github.com/swift-nav/piksi_apps/pull/563

PR #349, opened 2020-03-19
Fix clang-tidy

nsirola:
Clang-tidy did run at all. Tweak the settings so that it does and passes.

Note however that there is no `.clang-tidy` settings file in this project so it's run on quite lenient settings. Copying the settings from `libswiftnav` gave over 900 warnings, so there's some follow-up work there.


nsirola:
Looks like it does not get run:
```
[2020-03-19T07:35:53.064Z] + make clang-tidy-all
[2020-03-19T07:35:53.064Z] Scanning dependencies of target clang-tidy-all
[2020-03-19T07:35:53.064Z] Error: no input files specified.
[2020-03-19T07:35:53.064Z] USAGE: clang-tidy-6.0 [options] <source0> [... <sourceN>]
```

nsirola:
`clang-tidy` now gets run and found one issue:
image

nsirola:
`clang-tidy` runs now and gives green light.
Note however that there is no `.clang-tidy` settings file in this project so it's run on quite lenient settings. Copying the settings from `libswiftnav` gave over 900 warnings, so there's some follow-up work there.

c/CMakeLists.txt
```diff
@@ -20,7 +20,7 @@ if(CMAKE_C_COMPILER_ID MATCHES "GNU")
   # when using a GNU compiler. Doesn't apply when using clang compiler.
   set(clang_tidy_extra_args "EXTRA_ARGS" "-extra-arg=-Wno-unknown-warning-option")
 endif()
-swift_setup_clang_tidy(PATTERNS 'c/src/*.cc' ${clang_tidy_extra_args})
+swift_setup_clang_tidy(PATTERNS 'src/*.c' 'src/ubx_ephemeris/*.c' ${clang_tidy_extra_args})
```
anth-cole: why do we need to specify the `ubx_ephemeris` sub directory? isn't this covered in the first `src/*.c`?

nsirola: Ah true it is redundant, will fix before merging.

PR #351, opened 2020-03-24
Add .clang-tidy and implement fixes

nsirola:
Enable most of checks in `.clang_tidy`  and implement the suggested fixes. No functional changes intended.

Had to exclude only three checks:
- `hicpp-signed-bitwise` : would require every single `1 << 2` to be rewritten as `1U << 2U`
- `cert-err34-c`: requires all instances of `sscanf` to be replaced with `strtol` in rtcm3tosbp.c
- `llvm-include-order`: clashes with current `clang-format` settings

Test coverage takes a small hit because of many untested `else if` lines being split on two lines.


anth-cole:
lgtm - thanks for fixing.

PR #354, opened 2020-03-27
STAR-953: decode convert UBX GLO ephemeris [WIP]

nsirola:


c/src/ubx_ephemeris/glo.c
```diff
@@ -0,0 +1,166 @@
...
+static u8 reverse_byte(u8 b) {
+  b = (b & 0xF0) >> 4 | (b & 0x0F) << 4;
+  b = (b & 0xCC) >> 2 | (b & 0x33) << 2;
+  b = (b & 0xAA) >> 1 | (b & 0x55) << 1;
+  return b;
+}
+
+static gps_time_t glo2gps_wrapper(const glo_time_t *t) {
+  /* return time with factory UTC parameters */
+  /* TODO collect utc_params from UBX-MGA-GPS-UTC */
```
adel-mamin: The `TODO` note is still needed and to be addressed later I assume.

c/src/ubx_ephemeris/glo.c
```diff
@@ -0,0 +1,166 @@
...
+  return glo2gps(t, /* utc_params = */ NULL);
+}
+
+/**
+ * Decodes GLO pages 1-5.
+ * Reference: GLO OS SIS ICD, Issue 1.3, December 2016
+ * @param data context data
+ * @param prn transmitter's PRN
```
adel-mamin: `fcn` doc is missing.

c/src/ubx_ephemeris/glo.c
```diff
@@ -0,0 +1,166 @@
...
+  if (prn == UBX_UNKNOWN_PRN) {
+    /* PRN not yet known, don't try to decode */
+    return;
+  }
+
+  /* all pages collected, prepare the strings for decoder */
+
+  glo_string_t page[GLO_MAX_STRING_NUM];
+  assert(ARRAY_SIZE(page) == ARRAY_SIZE(sat->string));
+  memset(&page, 0, ARRAY_SIZE(page) * sizeof(glo_string_t));
```
adel-mamin: `memset(&page, 0, sizeof(page));`

c/src/ubx_ephemeris/glo.c
```diff
@@ -0,0 +1,166 @@
...
+
+  for (int i = 0; i < (int)ARRAY_SIZE(page); i++) {
```
adel-mamin: Maybe add a comment why this swap and reverse is needed.

c/src/ubx_ephemeris/glo.c
nsirola: Yes, or this will fail after the next leap second event (potentially 1st Jan 2021)

c/src/ubx_ephemeris/glo.c
nsirola: Good point. Noticed also that uBlox gives it as [0..13] while SBP expects [1..14] so fixed that.

c/src/ubx_ephemeris/glo.c
nsirola: Done. This seems a bit mad to me to reverse the bits twice, but this was the simplest rearrangement I could come up with to match the strings piksi firmware produces (It's possible that piksi firmware does things in a crazy order though)

c/src/ubx_ephemeris/glo.c
nsirola: Opened a follow-up item https://swift-nav.atlassian.net/browse/STAR-1136

c/src/ubx_ephemeris/glo.c
nsirola: I believe this is correct as the unit tests verify checksums of the produced strings, and also I successfully ran some UBX files converted with this through starling-runner. In case anyone wants to poke at this later, the unit tests should only pass if the resulting string was correct to the bit.

c/src/ubx_ephemeris/glo.c
jbangelo: @nsirola Mind including that ticket number in the TODO? In case it gets deprioritized and forgotten about.

c/src/ubx_ephemeris/glo.c
```diff
@@ -0,0 +1,168 @@
...
+         words,
+         sizeof(sat->string[string_num - 1].words));
+  if (ALL_STRINGS_MASK != (sat->vmask & ALL_STRINGS_MASK)) {
+    return; /* not all word types 1,2,3,4,5 are available yet */
+  }
+
+  if (prn == UBX_UNKNOWN_PRN) {
+    /* PRN not yet known, don't try to decode */
+    return;
+  }
```
jbangelo, line 112: It seems like this should be one of the first things we check in this function. We're using `prn` to index into array before this point, and it looks like we might be accessing out of bounds memory if this is true?

nsirola: Ouch, well spotted. Move this check to the very start.

c/src/ubx_ephemeris/glo.c
nsirola: Done

PR #363, opened 2020-04-02
Add cargo build wrapper for gnss converter tools [INFRA-151]

silverjam:
Requires https://github.com/swift-nav/libswiftnav/pull/129

Add deployment for gnss-converters to store pre-built copies of the tool on the GitHub releases page. Adds builds for Windows, macOS and Linux.  Release are available for every Git tag.

Adds a cargo build wrapper for gnss-converts to enable "two command" install:
```
git clone --recursive --jobs=4 https://github.com/swift-nav/gnss-converters.git
cargo +nightly install --path gnss-converters --bins
```

Uses nightly Rust to enable using `no_std` which allows the Rust wrapper to be as minimal as possible, but at the expense of requiring nightly Rust.


c/tests/check_rtcm3.c
```diff
@@ -1557,7 +1557,7 @@ static void compare_gps_ephs(const msg_ephemeris_gps_t *first,
   assert(first->common.sid.sat == second->common.sid.sat);
   assert(first->common.sid.code == second->common.sid.code);
   assert(first->common.toe.wn == second->common.toe.wn);
-  assert(fabs(first->common.toe.tow - second->common.toe.tow) <= 8);
+  assert((first->common.toe.tow - second->common.toe.tow) <= 8);
```
nsirola, line 1560: this is a bit strange in an ephemeris equality comparison function, I'll see separately if this can be replaced by an equality comparison as the values are integers

nsirola: https://github.com/swift-nav/gnss-converters/pull/365

silverjam: :+1: thanks @nsirola

PR #365, opened 2020-04-06
Use equality comparisons for integer tows in check_rtcm3

nsirola:
Ephemeris equality comparisons for some reason left 4-30 second slack for TOWs. Moreover after removing the surrounding `fabs` they allow unlimited error to the other direction. Fix by using equality comparisons as the TOWs are all integer seconds.

Turns out the example BDS ephemeris TOE and TOC values were invalid thus changed their value on round-trip conversion. Updated the example values.


c/tests/check_rtcm3.c
```diff
@@ -1653,7 +1653,7 @@ static msg_ephemeris_bds_t get_example_bds_eph() {
   ret.common.sid.sat = 25;
   ret.common.sid.code = CODE_BDS2_B1;
   ret.common.toe.wn = 2022;
-  ret.common.toe.tow = 460800;
+  ret.common.toe.tow = 460814;
```
nsirola, line 1656: The TOW values in SBP are in GPS time. The former value 460800 implied BDS timestamp of 460786 which is not divisible by 8 and thus is not a valid value for BDS ephemeris

PR #366, opened 2020-04-06
Update Libswiftnav

akleeman:
I was getting a fatal reference when updating submodules in orion engine which I think comes down to this change.

It seems to be referencing a commit from here: https://github.com/swift-nav/libswiftnav/pull/129 @silverjam ?


silverjam:
Weird, if this works, :shipit: ?

akleeman:
Yeah I can't figure out what happened, but updating libswiftnav to master seems to fix it :man_shrugging:

silverjam:
@akleeman I think I maybe had the submodule pointing at a branch that got deleted, for PR #363 I had gnss-converters pointing at a branch in libswiftnav, but probably didn't update the submodule pointer after the libswiftnav PR got merged

akleeman:
Ah that makes sense!  Oops.  Anyway it's trickling down the auto submodule chain so should be all clear!

martin-swift:
that's the usual suspect!

PR #367, opened 2020-04-07
Add ixcom2sbp to deploy

silverjam:
- Add ixcom2sbp to the deployed binaries for windows/linux/osx


silverjam:
@yizow defining `ixcom2sbp_main` allows the Rust wrapper to call `ixcom2sbp_main` as a library function and define it's own `main` entry-point for the Rust built version of program.

PR #369, opened 2020-04-08
[STAR-1136] use leap second from UBX-RXM-RAWX for GLO ephe conversion

nsirola:
SBP GLO ephemeris message has its timestamp in GPS time, and converting incoming GLO timestamps into GPS time requires up to date leap second information. This PR extracts that info from the UBX stream instead of relying on `libswiftnav`'s factory leap second table.

Updates submodules to their master tips to bring in:
- https://github.com/swift-nav/libswiftnav/pull/130
- https://github.com/swift-nav/libubx/pull/18

Tested with a 2-minute UBX log from which three rounds of GLO ephemeris are successfully decoded, and the leap second info from UBX-RMX-RAWX rather than the hard-coded table is applied.

Note that if ubx->sbp conversion is run over a leap second event, the receiving end should probably take some extra precautions, including at least either clearing the existing GLO ephemerides or updating their timestamps. The next round of ephemerides sent after the leap second event should get converted correctly again though.


c/src/ubx_sbp.c
```diff
@@ -338,14 +338,9 @@ void check_overflow(s32 *value) {
   }
 }
 
-static int fill_msg_obs(const u8 buf[], msg_obs_t *msg) {
```
nsirola, line 341: Refactored the decoding of `ubx_rxm_rawx` outside of this function

c/src/ubx_ephemeris/glo.c
```diff
@@ -87,6 +81,16 @@ void glo_decode_string(
...
   assert(data);
+
+  if (!data->leap_second_known) {
+    return;
+  }
+
+  utc_params_t utc_params;
+  memset(&utc_params, 0, sizeof(utc_params));
+  utc_params.dt_ls = data->leap_second;
+  utc_params.dt_lsf = data->leap_second;
```
jbangelo, line 92: Does setting just these fields actually cause the time conversion functions in libswiftnav to use the UTC params? I could have sworn there was a bit where we check the time of the leap second event in `utc_params_t` to make sure it's valid before using the leap second values.

c/src/ubx_ephemeris/glo.c
```diff
@@ -87,6 +81,16 @@ void glo_decode_string(
   }
 
   assert(data);
+
+  if (!data->leap_second_known) {
+    return;
```
jbangelo, line 86: Is there an issue with falling back to the hard coded leap second values in libswiftnav if we haven't received leap second info yet?

c/src/ubx_ephemeris/glo.c
nsirola: Uh, good point, it seems to revert to the hard-coded values if `wn==0`. The timestamp itself is only used to check if the `dt_ls` or `dt_lsf` value should be used.

c/src/ubx_ephemeris/glo.c
nsirola: Not as long as the hard coded value is correct, but once it falls behind the actual value it would produce 4km GLO ephemeris errors, potentially on every startup of the converter.

c/src/ubx_ephemeris/glo.c
nsirola: Fixed by storing the whole `utc_params` struct into the state

nsirola: And checked that the time conversion now actually goes into the branch using the utc parameters instead of the hard coded table

PR #375, opened 2020-04-17
Simply return if SBAS payload is faulty

scarcanague:


scarcanague:
@nsirola uBlox seems to output a null message for PRN138 (which is in test mode), leading to the assert

PR #379, opened 2020-04-23
[PMF-88] Fix the orientation angle in GPGST

nsirola:
Fix the division by zero in GPGST (https://swift-nav.atlassian.net/browse/PMF-88) that happens if solution covariance should be a zero matrix, or if the covariance's semimajor axis points straight East. 

Also, the print out the orientation angle in degrees instead of radians:
image


nsirola:
It came up in the UAT tests

PR #381, opened 2020-04-28
ESD-1784: gnss-converters must be platform agnostic 

RReichert:
As part of the `swiftlets` work is to make sure that any dependencies operate in a platform agnostic manner or at least redirect all platform specific operations through the `libpal` library. Seeing as swiftlets` uses `gnss-converters` extensively, a few changes needed to be made:

Within the library, based off of what I can tell, there are two platform specific operations in use:

* memory allocation via `malloc`
* streaming content out to standard stream

The solution to the memory allocation is to simply move that bit of code outside of the `gnss-converters` library into a new library `gnss-converters-extra`, seeing as the work really was a wrapper around the library from what I heard. This will affect orion so as part of this work I'll have to make sure everything works there before merging.

As for the heavy use of the standard output stream, I've redirected all those error reporting through the `swiftnav` logging framework. I've assumed that each of the `printf` statements were mostly for reporting purposes rather than major error, which is why everything was converted over to using `log_info` over any of the other logging levels. If this is problematic, please let me know.  

As part of the fix, a bit of restructuring was made (not necessary) to clearly isolate what code belonged to what target, hopefully also making things a lot more manageable.


c/src/gnss-converters/rtcm3_sbp.c
```diff
@@ -17,9 +17,23 @@
...
+#ifdef _WIN32
+#ifdef _WIN64
+#define PRI_SSIZE "I64d"
+#else
+#define PRI_SSIZE "I32d"
+#endif
+#else
+#define PRI_SSIZE "zd"
+#endif
+
```
RReichert, line 34: I know the comment above says that the fix was suppose to correct the `%zd` issue, but that didn't seem to work when shifting over to swiftnav's logging framework function. maybe there is a stronger set of validations put in place for it.

c/src/gnss_converters/ubx_sbp.c
```diff
@@ -184,12 +185,12 @@ static int read_ubx_frame(u8 *frame, struct ubx_sbp_state *state) {
      * bytes for length, 2 bytes for checksum not counted in payload_length
      */
     if (payload_length > UBX_FRAME_SIZE - 6) {
-      fprintf(stderr,
-              "UBX payload_length for class 0x%X and ID 0x%X too large: %d; "
-              "possible corrupted frame\n",
-              frame[0],
-              frame[1],
-              payload_length);
+      log_info(
```
pgrgich, line 188: This one should probably be more than an info.

c/src/gnss_converters/ubx_sbp.c
```diff
@@ -454,9 +455,8 @@ static int fill_msg_pos_llh(const u8 buf[], msg_pos_llh_t *msg) {
           msg->flags |= SBP_LLH_FIXED_RTK_MASK;
           break;
         default:
-          fprintf(stderr,
-                  "Invalid NAV_PVT.flags carrier solution value: %d\n",
-                  carr_soln);
+          log_info("Invalid NAV_PVT.flags carrier solution value: %d",
```
pgrgich, line 458: This should probably also be higher than an info.

c/src/gnss_converters/rtcm3_sbp.c
```diff
@@ -125,7 +140,7 @@ void rtcm2sbp_decode_payload(const uint8_t *payload,
       (payload[byte] << 4) | ((payload[byte + 1] >> 4) & 0xf);
 
   if (verbosity_level > VERB_HIGH) {
-    fprintf(stderr, "MID: %5u\n", message_type);
+    log_info("MID: %5u", message_type);
```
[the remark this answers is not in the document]
silverjam: Never mind, the `logging.h` is a bit odd in general, by default all logging is enable except for LOG_DEBUG, and it seems there's no way to change that in upstream libraries without recompiling.

c/src/gnss_converters/ubx_sbp.c
RReichert: done

c/src/gnss_converters/ubx_sbp.c
RReichert: done

PR #390, opened 2020-05-10
[SEN-268] Fix temperature field scaling

reimerix:
This PR fixes the temperature field scaling for the IMU temperature.﻿


c/src/gnss_converters/ubx_sbp.c
```diff
@@ -801,7 +801,7 @@ static void send_imu_aux(struct ubx_sbp_state *state) {
   const u8 acc_range_4g = 1;
   msg.imu_conf = (gyro_range_125degs << 4) | acc_range_4g;
   double imu_temp_bmi160_scaled =
-      (state->esf_state.last_imu_temp - 23.0) / 512.0;
+      (state->esf_state.last_imu_temp - 23.0) * 512.0;
```
scarcanague, line 804: I was wondering, why not encoding in Celcius? If we need to match an existing implementation (BMI?) maybe put a reference to the BMI datasheet to explain why you pick this encoding?

reimerix: Since the rest of our implementation uses BMI160 encoding, we wanted to do the same thing here. I added comments with a link to the datasheet and some unit tests for the conversion to make sure we encode correctly.

PR #398, opened 2020-05-26
[SSR-95] Expand IODE to 16 bits

ljbade:


ljbade:
I'll remove that TODO on the BDS IODE and make a bug ticket for it

c/src/gnss_converters/sbp_rtcm3_ephemeris.c
```diff
@@ -256,8 +256,8 @@ void sbp_to_rtcm3_gal_eph(const msg_ephemeris_gal_t *sbp_gal_eph,
   msg_eph->kepler.af1 = (int32_t)(round(sbp_gal_eph->af1 / C_1_2P46));
   msg_eph->kepler.af2 = (int16_t)(round(sbp_gal_eph->af2 / C_1_2P59));
 
-  msg_eph->kepler.iode = sbp_gal_eph->iode;
-  msg_eph->kepler.iode = sbp_gal_eph->iodc;
```
seth-swiftnav, line 260: nice fix.

seth-swiftnav, line 260: Nice fix.

seth-swiftnav, line 260: Nice fix

c/src/gnss_converters/sbp_rtcm3_ephemeris.c
```diff
@@ -199,8 +199,8 @@ void sbp_to_rtcm3_bds_eph(const msg_ephemeris_bds_t *sbp_bds_eph,
   msg_eph->kepler.af1 = (int32_t)(round(sbp_bds_eph->af1 / C_1_2P50));
   msg_eph->kepler.af2 = (int16_t)(round(sbp_bds_eph->af2 / C_1_2P66));
 
-  msg_eph->kepler.iode = sbp_bds_eph->iode;
-  msg_eph->kepler.iodc = sbp_bds_eph->iodc;
+  msg_eph->kepler.iode = 0;
+  msg_eph->kepler.iodc = 0;
```
pgrgich, line 203: Why 0? Don't we want to make it the RTCM value?

ljbade: We don't write these values out since the RTCM BDS message has no IODE/IODC fields at all. This was why AODE was being stuffed into here for the round trip. I can drop this change if we follow it up when AODE bug is fixed.

PR #402, opened 2020-05-28
Upgrade to Xcode 10, pin Rust nightly [INFRA-202]

silverjam:
Attempting to resolve build errors here: https://travis-ci.org/github/swift-nav/gnss-converters/jobs/692078797

Something is wonky with xcode8 nodes on Travis... xcode8 is from 2016, so we can just bump to xcode10 (release in 2018) to fix this issue.  Xcode 10 on Travis is macOS version 10.12.


.travis.yml
```diff
@@ -39,7 +39,7 @@ matrix:
 
     - os: linux
       language: rust
-      rust: nightly
+      rust: nightly-2020-05-25
```
jbangelo, line 42: Any particular reason to use this specific version of nightly?

silverjam: No, it's just the latest one it built with, so pinning with the last version that worked.

PR #406, opened 2020-05-30

dgburr:
The current implementation of pack_carrier_phase() and pack_doppler() incorrectly rounds values which are very close to a whole number.  For example, 1613.994385 should be represented as 1613/255, not 1614/0.


dgburr:
It appears that this code has been duplicated into/copied from starling (https://github.com/swift-nav/starling/blob/master/src/util/sbp/packers.c#L482) but the bug has already been fixed in the starling version.  So it seems that the proper solution is to refactor these functions to use a common (correct) implementation.

c/src/gnss_converters/ubx_sbp.c
```diff
@@ -291,7 +291,7 @@ static bool pack_carrier_phase(double L_in, carrier_phase_t *L_out) {
 
   L_out->i = (s32)Li;
   u16 frac_part_cp = (u16)round(Lf * SBP_OBS_LF_MULTIPLIER);
-  if (frac_part_cp >= UINT8_MAX) {
+  if (frac_part_cp >= 256) {
```
jbangelo, line 294: Could we instead have this? It makes the intension slightly more obvious IMO.

```suggestion
  if (frac_part_cp > UINT8_MAX) {
```

dgburr: Done, but as mentioned above, the same functionality exists in multiple places (https://github.com/swift-nav/starling/blob/master/src/util/sbp/packers.c#L482, https://github.com/swift-nav/gnss-converters/blob/master/c/src/gnss_converters/rtcm3_sbp.c#L839) and different logic is used each time.

dgburr: There seems to be no easy way to rebase this change onto starling-v1.1.0-release - @jbangelo, should I just open a new PR instead?

dgburr: Merged to starling-v1.1.0-release in #413. The remaining part of this PR (i.e. to leave the setting of the `valid` flag to `decode_gal_ephemeris()`) is dependent upon https://github.com/swift-nav/libswiftnav/pull/144

dgburr:
Closed since this change has now been merged to master in https://github.com/swift-nav/gnss-converters/pull/531

PR #413, opened 2020-06-11

dgburr:
The current implementation of pack_carrier_phase() and pack_doppler() incorrectly rounds values which are very close to a whole number. For example, 1613.994385 should be represented as 1613/255, not 1614/0.


dgburr:
This PR transfers the change to #406 to the release branch

PR #414, opened 2020-06-15
Fix GLO ephemeris decoding crash.

pgrgich:
Fix for https://swift-nav.atlassian.net/browse/POS-197. It will also prevent forwarding of a GLO ephemeris before we have the leapsecond (and therefore have an invalid time). Presumably this is desirable as the only way we could make use of an ephemeris with invalid time would be if the RTCM time conversion from `t_b` is done downstream or `starling` can make use of an ephemeris with invalid time. Neither seems plausible to me.


pgrgich:
Verified that it fixed the problem. @mfine any concerns?

PR #416, opened 2020-06-30
Add AODE/AODC to BDS ephs.

pgrgich:
One part of a change that will involve adding a new SBP message for BDS ephs. The problem is that the current one round trips by storing the AODE/AODC in the IODE/IODC parameters. This is confusing and we also want to use the IODE/IODC as a unique key (in a later PR).


pgrgich:
@ljbade yes it will. I'll get that in once I've figured out how it should be done. :)

PR #441, opened 2020-07-23
[HS-60] Output CPU local timestamps for IMU and wheelspeed measurements

reimerix:
This PR changes the time reference for IMU and wheelspeed output to CPU local time. Instead of directly translating the CPU local time to GNSS time, `ubx2sbp` will now always output in CPU local time and additionally send out a `MSG_GNSS_TIME_OFFSET` if available.

While trying to chase down some CI-only failures, I compiled with `-fsanitize=undefined` which found unrelated issues - fixed them in 64d2cd3 .

@scarcanague : Could you please review the fix to `convert_meters_to_sisa `?


c/src/gnss_converters/rtcm3_utils.c
```diff
@@ -935,6 +935,10 @@ float convert_sisa_to_meters(const uint8_t sisa) {
 
 uint8_t convert_meters_to_sisa(const float ura) {
   /* Convert between meters and GAL SISA index.*/
+  if (ura < 0) {
```
reimerix, line 938: This function would invoke undefined behavior if `ura < 0` before (this didn't fall through to the last line). It will now output "No accuracy prediction available" instead.

c/src/gnss_converters/sbp_nmea.c
```diff
@@ -87,7 +87,11 @@ struct sbp_meta_entry {
 };
 
 static uint32_t get_tow(const sbp2nmea_t *state, sbp2nmea_sbp_id_t id) {
-  return *(uint32_t *)(sbp2nmea_msg_get(state, id) + sbp_meta[id].offset_tow);
```
reimerix, line 90: This would invoke undefined behavior because we were loading with a misaligned a address before. Doing a memcpy into a local variable instead fixes this issue.

jbangelo: Good find!

reimerix:
@jbangelo There are some reasons to go with representing IMU timestamps as `gps_time_t` in the context of converting to SBP that I‘d like to point out:

- `MSG_IMU_RAW` only has 30 bits left for actual IMU timestamps, as the upper two bits are flags to represent the timing mode. Thus, no „natural“ 32 bit overflow to zero can be represented.
-  The SBP spec demands that IMU timestamps wrap around to zero after one week if using CPU local timestamps: One week, represented as milliseconds, would need 29.17 bits. Instead of going with an (at least to me) equally arbitrary overflow of 30 bits, going with `gps_time_t` enables us to use all the functions related to that and thus greatly eases handling the data.

The  `msss` field provided by the ubx NAV-STATUS message gets converted to s64 from an u32 because we don’t need to handle overflows in a s64 (it only wraps around after 293 million years). We initialize the s64 with the first value we receive and afterwards only update it by diffs of the u32 input values.

Do you think providing these bits of informations as comments in the source code would be an acceptable compromise?

c/src/gnss_converters/rtcm3_utils.c
scarcanague: Good catch. We probably want this here too: https://github.com/swift-nav/gnss-converters/pull/441/commits/64d2cd36e749e099e9305dff0ed9c7a787686e0d#diff-28e759a88c64afcf5888a8937f86c44fR958

c/src/gnss_converters/ubx_sbp.c
```diff
@@ -561,16 +561,28 @@ static u8 get_odo_velocity_source(u8 data_type) {
   }
 }
 
-static void set_wheeltick_time(u32 msss, msg_wheeltick_t *msg_wheeltick) {
+static void set_wheeltick_time(s64 msss, msg_wheeltick_t *msg_wheeltick) {
```
scarcanague, line 564: Do we need signed integer if it is a counter?

c/src/gnss_converters/ubx_sbp.c
```diff
@@ -561,16 +561,28 @@ static u8 get_odo_velocity_source(u8 data_type) {
...
   const u8 time_source_local = 2;
   msg_wheeltick->flags = time_source_local;
-  msg_wheeltick->time = (u64)(msss * SECS_US / SECS_MS);
+  // The SBP protocol expects CPU local timestamps to wrap around after one
```
scarcanague, line 567: Can't the SBP protocol handle system time with the type of timestamp flags set to 2?

c/src/gnss_converters/ubx_sbp.c
```diff
@@ -748,46 +714,36 @@ static void maybe_parse_imu_data(u32 data,
...
-  sensortime -= ubx_imu_gnss_time_offset;
-
-  gps_time_t gps_sensortime;
-  gps_sensortime.tow = sensortime;
-  gps_sensortime.wn = 0;
-  unsafe_normalize_gps_time(&gps_sensortime);
-  sensortime = gps_sensortime.tow;
-
-  struct sbp_imuraw_timespec timespec = convert_tow_to_imuraw_time(sensortime);
+  gps_time_t imu_time_msss;
```
scarcanague, line 763: Yeah agree with Joseph, it is a bit confusing to fit the sensor time into a GPS time structure. Isnt the SBP handling a number of milliseconds with unknown reference epoch?

c/src/gnss_converters/ubx_sbp.c
reimerix: The `running_msss` fields in `ESF-STATE`  are `s64` (mainly to be able to denote an invalid initial value of -1) and they get passed into this function. Shouldn't matter given that we already have enough space for 293 million years in 63 bits.

c/src/gnss_converters/ubx_sbp.c
reimerix: Tried to explain that choice above: 

> MSG_IMU_RAW only has 30 bits left for actual IMU timestamps, as the upper two bits are flags to represent the timing mode. Thus, no „natural“ 32 bit overflow to zero can be represented.
The SBP spec demands that IMU timestamps wrap around to zero after one week if using CPU local timestamps: One week, represented as milliseconds, would need 29.17 bits. Instead of going with an (at least to me) equally arbitrary overflow of 30 bits, going with gps_time_t enables us to use all the functions related to that and thus greatly eases handling the data.

To be totally clear, the constraint of wrapping around to zero after 604800 seconds instead of 1073741.824 seconds (30 bits with millisecond resolution) is an arbitrary choice - but I think it makes more sense to handle it this way rather than re-implementing all the wrap-around, time matching and arithmetic code for 30 bit overflows on millisecond time tags. We are constrained to those 30 bits (in millisecond resolution) because we don't want to have different time tags (in `MSG_IMU_RAW`, `MSG_ODOMETRY`, `MSG_WHEELTICK`) representing the same time and we only have 30 bits left to use in the TOW field of `MSG_IMU_RAW`.

FWIW, I agree that the name being `gps_time_t` is somewhat confusing - I feel like that is the main issue here?

c/src/gnss_converters/ubx_sbp.c
reimerix: It can, but since `MSG_IMU_RAW` is constrained to 30 bits with millisecond resolution, using a rollover after a week didn't seem like a very bad choice. I think we would definitely want IMU message timestamps to align nicely with wheeltick or odometry timestamps if they are received directly after each other.

reimerix:
After syncing with @jbangelo, the plan is to accept this PR as it stands and to revisit this and other internal pieces concerned with IMU local timestamps in the following months. A ticket for this has been added into our backlog.

PR #462, opened 2020-08-15
Disable uBlox ephemerides.

pgrgich:


pgrgich:
We fixed the bug for real instead. :)

PR #463, opened 2020-08-17
Fix ubx2sbp conversion bug.

pgrgich:
It looks like this was the problem all along...


c/include/gnss-converters/ubx_sbp.h
```diff
@@ -40,6 +40,11 @@ struct sat_data {
   unsigned vmask;
 };
 
+inline void invalidate_subframes(struct sat_data *sat, unsigned mask) {
```
pgrgich, line 43: Unify this.

c/src/gnss_converters/ubx_ephemeris/bds.c
```diff
@@ -103,6 +98,7 @@ void bds_decode_subframe(struct ubx_sbp_state *data,
   }
 
   if ((sow[0] != (sow[1] - 6)) || (sow[1] != (sow[2] - 6))) {
+    invalidate_subframes(sat, /*mask=*/7);
```
pgrgich, line 101: In this case, are all the previously read in subframes out of date and the current one valid or do we want to throw them all out? I would guess the former, but I've implemented the latter until I get confirmation because of the risk.

c/src/gnss_converters/ubx_ephemeris/gal.c
```diff
@@ -126,6 +126,7 @@ void gal_decode_page(struct ubx_sbp_state *data,
 
   for (int i = 1; i < (int)ARRAY_SIZE(iod); i++) {
     if (iod[i] != iod[0]) {
+      invalidate_pages(sat, /*mask=*/ALL_PAGES_MASK & ~(1U << (wtype - 1)));
```
pgrgich, line 129: Discard everything but the current subframe (this is `wtype - 1` for some reason).

c/src/gnss_converters/ubx_ephemeris/gps.c
```diff
@@ -100,7 +95,7 @@ void gps_decode_subframe(struct ubx_sbp_state *data,
   u8 iode8_sf3 = (sat->sf[2].words[9] >> 22U) & 0xFFU;
 
   if ((iodc8_sf1 != iode8_sf2) || (iode8_sf2 != iode8_sf3)) {
-    invalidate_subframes(sat);
+    invalidate_subframes(sat, /*mask=*/0x7 & ~(1U << sf_idx));
```
pgrgich, line 98: Discard everything but the current subframe (this is `sf_idx`).

c/src/gnss_converters/ubx_ephemeris/gps.c
```diff
@@ -112,7 +107,7 @@ void gps_decode_subframe(struct ubx_sbp_state *data,
   } else if (tow_6s < WEEK_SECS / 6) {
     tot_tow_s = (tow_6s - 1) * 6;
   } else {
-    invalidate_subframes(sat);
+    invalidate_subframes(sat, /*mask=*/0x7);
```
pgrgich, line 110: What is this case? Do we want to keep the current subframe?

pgrgich: I think this if `tow_6s` is invalid (multiplying by 6 gives > `WEEK_SECS`).

c/src/gnss_converters/ubx_ephemeris/gal.c
```diff
@@ -115,21 +115,29 @@ void gal_decode_page(struct ubx_sbp_state *data,
...
-  if (ALL_PAGES_MASK != (sat->vmask & ALL_PAGES_MASK)) {
-    return; /* not all word types 1,2,3,4,5 are available yet */
-  }
 
-  int iod[4];
-  for (int i = 0; i < (int)ARRAY_SIZE(iod); i++) {
-    iod[i] = (words[0] >> 14U) & 0x3FFU;
+  // Get the stored IODnavs for the first four subframes - the fifth has no
+  // IODnav
+  int IODnav[4];
```
ljbade, line 121: I would make this a `u16`

c/src/gnss_converters/ubx_ephemeris/gal.c
```diff
@@ -115,21 +115,29 @@ void gal_decode_page(struct ubx_sbp_state *data,
...
+  for (int i = 0; i < (int)ARRAY_SIZE(IODnav); i++) {
```
ljbade, line 122: if this was a `size_t i` then you remove need for the cast

c/src/gnss_converters/ubx_ephemeris/gal.c
```diff
@@ -115,21 +115,29 @@ void gal_decode_page(struct ubx_sbp_state *data,
...
+    IODnav[i] = (sat->pg[i].words[0] >> 14U) & 0x3FFU;
   }
 
-  for (int i = 1; i < (int)ARRAY_SIZE(iod); i++) {
-    if (iod[i] != iod[0]) {
-      return; /* received pages are not from the same batch */
+  // If we haven't just read in the fifth subframe, check to see if this
+  // subframe is newer than the old ones
+  if (wtype - 1 != 4) {
+    int new_IODnav = (words[0] >> 14U) & 0x3FFU;
```
ljbade, line 129: `u16`

c/src/gnss_converters/ubx_ephemeris/gal.c
```diff
@@ -115,21 +115,29 @@ void gal_decode_page(struct ubx_sbp_state *data,
...
+    for (int i = 0; i < (int)ARRAY_SIZE(IODnav); i++) {
```
ljbade, line 130: `size_t i`

c/src/gnss_converters/ubx_ephemeris/gal.c
```diff
@@ -115,21 +115,29 @@ void gal_decode_page(struct ubx_sbp_state *data,
...
+  for (int i = 0; i < (int)ARRAY_SIZE(IODnav); i++) {
+    IODnav[i] = (sat->pg[i].words[0] >> 14U) & 0x3FFU;
   }
 
-  for (int i = 1; i < (int)ARRAY_SIZE(iod); i++) {
-    if (iod[i] != iod[0]) {
-      return; /* received pages are not from the same batch */
+  // If we haven't just read in the fifth subframe, check to see if this
+  // subframe is newer than the old ones
+  if (wtype - 1 != 4) {
```
ljbade, line 128: personally id use `wtype < 5` to match the comment better

c/src/gnss_converters/ubx_ephemeris/gps.c
ljbade: might be worth making this a code comment ^

c/src/gnss_converters/ubx_ephemeris/gps.c
ljbade: probably only need to invalidate subframe 0?

tao-swift:
RTKLIB has some similar work decoding the BDS, GAL raw navigation message, for instance, the decoding of BDS D1 is given in https://github.com/tomojitakasu/RTKLIB/blob/71db0ffa0d9735697c6adfd06fdf766d0e5ce807/src/rcv/ublox.c#L851, with the function definition in https://github.com/tomojitakasu/RTKLIB/blob/71db0ffa0d9735697c6adfd06fdf766d0e5ce807/src/rcvraw.c#L202. The `+6` is also used in here https://github.com/tomojitakasu/RTKLIB/blob/71db0ffa0d9735697c6adfd06fdf766d0e5ce807/src/rcvraw.c#L257. Hopefully those can help us cross validate our code.

c/src/gnss_converters/ubx_ephemeris/gps.c
pgrgich: Good idea.

c/src/gnss_converters/ubx_ephemeris/gps.c
pgrgich: Yep.

c/src/gnss_converters/ubx_ephemeris/gal.c
pgrgich: Yep, that's better.

pgrgich:
@ljbade agreed on the `int`s and so forth, but I think we should change all four files (gps.c, glo.c, bds.c and gal.c) at the same time so that we retain consistency and that's too much for this PR. :)

PR #473, opened 2020-08-24
[RC-224] Fix temperature encoding for negative temperatures

reimerix:
We were using a wrong temperature encoding because we were using the full 32 bit "data" value instead of the 24 bit portion of it containing the actual signed data.

Furthermore, the change introducing CPU local timestamps (8cf75ed4c0ef4bdfb75e39a579d0e0b229a47262) introduced a bug which would lead to never decoding the IMU temperature from the input data stream.

This PR fixes both the encoding and output problem and adds unit tests for the temperature encoding.

Also fixes [SEN-392].

[SEN-392]: https://swift-nav.atlassian.net/browse/SEN-392


c/tests/check_ubx.c
```diff
@@ -649,6 +649,40 @@ static void ubx_sbp_callback_test_no_conversion_invalid_calibtag(
...
+  uint16_t expected_messages[] = {SBP_MSG_IMU_AUX, SBP_MSG_IMU_RAW};
+  ck_assert_int_eq(msg_id, expected_messages[msg_index]);
+
+  if (msg_index == 0) {
+    msg_imu_aux_t *msg = (msg_imu_aux_t *)buff;
+    ck_assert_int_eq(msg->temp, (30 - 23) * 512);
+  }
+
+  msg_index++;
+}
```
richarddeurloo, line 684: Nit: I know this a test, but the above is mostly duplicate code.

richarddeurloo (commented):
Just some minor comments

c/src/gnss_converters/ubx_sbp.c
```diff
@@ -855,6 +847,13 @@ static void handle_esf_raw(struct ubx_sbp_state *state, u8 *inbuf) {
   s16 current_msg_number = 0;
   for (int i = 0; i < num_raw; i++) {
     u8 data_type = (esf_raw.data[i] >> 24) & 0xFF;
+    u32 data = esf_raw.data[i] & 0xFFFFFF;
+    if (data_type == ESF_GYRO_TEMP) {
+      s32 parsed_value = convert24b(data);
+      const double ubx_scale_temp = 0.01;
```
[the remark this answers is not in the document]
reimerix: unfortunately not possible since this is C code

c/tests/check_ubx.c
reimerix: Valid point. I changed it to a single function which uses the context argument instead.

PR #478, opened 2020-08-28
[SR-52] Most sentences not output in DR mode

woodfell:
When in DR mode with no GNSS fix PVT most useful NMEA sentences are not written. They depend on SBP input messages AGE_CORRECTIONS and DOPS which are not being output by PVT Engine because no fix is available, and are not output by fusion engine. Fusion engine continues to output dead reckoning solutions but they are not converted to NMEA.

The set of SBP input messages required to output a particular sentence is controlled by a mask in sbp2nmea. This mask defines which SBP input messages are require for each NMEA output sentence. This mask is checked every time a new input message arrives and if every required input message is valid the sentence is written. In DR mode this condition is never fulfilled.

This PR splits the mask in to 2 parts. The first mask (`base_tow_mask`) is applied in all situations. The second (`non_dr_tow_mask`) is applied only when the position mode in `pos_llh_cov` is not dead reckoning. This allows all sentences to be written out even when only DR solutions are available.

An additional change to the GPGGA sentence to output a null field for HDOP when in DR mode. Age corrections field is already null in DR mode.


woodfell:
I don't know what's going on with Travis, it's been ages and it hasn't even start the build yet. Going to merge anyway since previous versions built ok and it's only had a rebase since then

PR #486, opened 2020-09-01
[SR-85]: Accept UBX-SFRBX messages with length == 9 in order to work around an apparent bug in the HPG1.13 F9P firmware

dgburr:
The extra word appears to be unused since the CRC check (which only checks the first 250 bits) still works as expected with these messages.


dgburr:
Created https://github.com/swift-nav/gnss-converters/pull/491 with the same fix ported to starling-v1.2.0-release

PR #489, opened 2020-09-03
Add UBX MON HW forwarding

benjaminaltieri:


reimerix (approved):
LGTM

JamesTidd (approved):
Changes look good.  Should there be some changes to Swiftlets YAML config files for specific customers to route the data from ME input to Starling (the product, not the repo) output, or are we not controlling these in GitHub?

benjaminaltieri:
> Changes look good. Should there be some changes to Swiftlets YAML config files for specific customers to route the data from ME input to Starling (the product, not the repo) output, or are we not controlling these in GitHub?

There is no dynamic control at this point, this simply hardcodes the message that is desired and forwards it.

PR #494, opened 2020-09-08
Support fusion engine output in sbp2nmea

woodfell:


c/include/gnss-converters/sbp_nmea.h
```diff
@@ -63,19 +65,30 @@ typedef struct sbp_state_entry {
   sbp2nmea_msg_t msg;
 } sbp_state_entry_t;
 
+typedef enum sbp2nmea_mode {
+  SBP2NMEA_MODE_BEST,
+  SBP2NMEA_MODE_GNSS,
+  SBP2NMEA_MODE_FUZED,
```
benjaminaltieri, line 71: Let's avoid the 'z', although there seems to be some uptake in this alternative spelling: fusion, fuse and fused are more directly recognizable as correct spelling/conjugation

c/include/gnss-converters/sbp_nmea.h
```diff
@@ -111,8 +127,6 @@ void sbp2nmea_msg_length_set(sbp2nmea_t *state, u8 len, sbp2nmea_sbp_id_t id);
 
 void sbp2nmea_rate_set(sbp2nmea_t *state, int rate, sbp2nmea_nmea_id_t id);
 
-void sbp2nmea_soln_freq_set(sbp2nmea_t *state, float soln_freq);
```
benjaminaltieri, line 114: This may have implications on the use case in Piksi, updating the solution frequency while running may needs to be a supported use case but I can't remember exactly how it is used atm

c/src/gnss_converters/sbp_nmea.c
```diff
@@ -26,56 +26,50 @@
...
-                                            (1 << SBP2NMEA_SBP_DOPS) |
-                                            (1 << SBP2NMEA_SBP_AGE_CORR),
-                           .non_dr_tow_mask = 0,
+    [SBP2NMEA_NMEA_ZDA] = {.tow_mask = 0, .send = send_gpzda},
+    [SBP2NMEA_NMEA_GSA] = {.tow_mask = (1 << SBP2NMEA_SBP_POS_LLH_COV) |
+                                       (1 << SBP2NMEA_SBP_GPS_TIME) |
+                                       (1 << SBP2NMEA_SBP_VEL_NED) |
+                                       (1 << SBP2NMEA_SBP_DOPS) |
+                                       (1 << SBP2NMEA_SBP_AGE_CORR),
+
```
benjaminaltieri, line 64: Whitespace

c/src/gnss_converters/sbp_nmea.c
```diff
@@ -113,63 +109,96 @@ static double sbp_gpsdifftime(const sbp_gps_time_t *sbp_end,
...
+    freq = state->gnss_soln_freq;
+  } else {
+    if (!state->fuzed_wagon_complete) {
+      return;
+    }
+
+    msg_group_meta_t *group_meta =
+        sbp2nmea_msg_get(state, SBP2NMEA_SBP_GROUP_META);
+
+    if (group_meta->group_id == 2) {
```
benjaminaltieri, line 196: Is this actually functional yet? Nice to have a call out for the future, would be better to use a define that indicate what 2 means here (implicit from the resulting execution but might not be for others in the future)

c/src/gnss_converters/sbp_nmea.c
```diff
@@ -206,10 +235,83 @@ static bool sbp2nmea_discard_sbp(const sbp2nmea_sbp_id_t sbp_id,
...
+
+  return SBP2NMEA_SBP_CNT;
+}
+
 void sbp2nmea(sbp2nmea_t *state,
               const u8 len,
               const void *sbp_msg,
-              sbp2nmea_sbp_id_t sbp_id) {
+              uint16_t msg_type) {
+  if (msg_type == SBP_MSG_GROUP_META) {
```
benjaminaltieri, line 298: It would be helpful to have a bit more structure to this block, there a couple things going on including checking for complete wagons and switching modes.  A clear definition of how the fusion wagon is determined complete looks now to be has a solution meta message been received following a group meta message. My understanding (this may be for future improvement) was that the group meta had everything needed to verify proper receipt of the group but it isn't used here.

c/src/gnss_converters/sbp_nmea.c
```diff
@@ -206,10 +235,83 @@ static bool sbp2nmea_discard_sbp(const sbp2nmea_sbp_id_t sbp_id,
...
+    state->fuzed_wagon_complete = false;
+    if (state->requested_mode == SBP2NMEA_MODE_BEST &&
```
benjaminaltieri, line 300: Should we make sure the group meta is not a GNSS group, as implemented above?

benjaminaltieri (changes requested):
This is great, looks exactly like the potential changes the output refactor for the fusion engine would enable. Ironically (or not) main bone to pick here is on naming/spelling which I would prefer not to canonize. Otherwise would have approved. Rest of the comments questions are there for your consideration. Overall awesome work on this :+1:

c/include/gnss-converters/sbp_nmea.h
woodfell: Oh I'm so glad someone else said that :smile: I though 'z' was an American thing and I was just going to have to go with it

c/include/gnss-converters/sbp_nmea.h
woodfell: Yeah now you mention it it does need to be updated in real time. I'll put the "set" functions back in

c/src/gnss_converters/sbp_nmea.c
woodfell: No idea. I saw it in the struct definition so went with it. Haven't seen it actually be used in testing though. I'll get rid of it for the moment.

c/include/gnss-converters/sbp_nmea.h
benjaminaltieri: I was thinking of the irony, that this is the one 'z' you didn't rebel against :joy:

c/src/gnss_converters/sbp_nmea.c
woodfell: Reworked it a little, split some stuff out to separate functions. Hopefully reads a bit better now. 

Also added some state information to keep track of the fused wagon. It will be marked as complete once all messages have been received rather than when SOLN_META is received (although in practice these are the same). Waiting to hear back about the contents of `group_msgs` in `GROUP_META` - the comments in libsbp suggest that `GROUP_META` itself should be a member of that array but in the SBP log I'm testing with it isn't.

c/src/gnss_converters/sbp_nmea.c
woodfell: Is this still relevant?

anth-cole (approved):
lgtm

PR #497, opened 2020-09-10
Output PUBX from sbp2nmea

woodfell:
Adds PUBX as an output of sbp2nmea

Calculating Vacc and Hacc as per this comment https://github.com/swift-nav/gnss-converters/pull/497#pullrequestreview-485562583

DOPS are null when in DR mode


anth-cole (approved):
lgtm - I took a look at the POS_LLH_COV message and it seems we already have the cov values in the east/north/down reference frame.

So I think the Hacc field should be sqrt(cov_n_n^2+cov_e_e^2) and the Vacc field should just be cov_d_d - can you confirm @scarcanague or @ericsenant that that follows what you would expect?

benjaminaltieri (approved):
LGTM

c/src/gnss_converters/nmea.c
```diff
@@ -1105,6 +1105,137 @@ void send_gsv(const sbp2nmea_t *state) {
...
+
+  // NavStat
+  const char *nav_stat = get_pubx_nav_stat(sbp_pos_llh_cov->flags);
+  NMEA_SENTENCE_PRINTF("%s,", nav_stat);
+
+  // Hacc, Vacc
+  if (fix_type != NMEA_GGA_QI_INVALID) {
+    float hacc = sqrtf(sbp_pos_llh_cov->cov_n_n * sbp_pos_llh_cov->cov_n_n +
+                       sbp_pos_llh_cov->cov_e_e * sbp_pos_llh_cov->cov_e_e);
+    float vacc = sbp_pos_llh_cov->cov_d_d;
```
scarcanague, line 1187: We should `sqrtf` that if result is in meters?

woodfell: I have to defer to someone else here @anth-cole @ericsenant 

If `cov_n_n` and `cov_e_e` are in `m^2` does that mean hacc should be just `sqrtf(cov_n_n + cov_e_e)` instead?

pgrgich: I'd say so, assuming that we want the RMS to be in the `hacc` variable (we define it as the norm in SBP, which is different). `vacc` should also be the `sqrtf`, right?

ericsenant: yes, to me sqrtf(cov_n_n + cov_e_e) is what should be done as cov_n_n and cov_e_e are already squared sigmas on East and North
the same way I would think that vacc = sqrtf(cov_d_d)

woodfell: Ok so are we settled on 
```
hacc = sqrtf(cov_n_n + cov_e_e)
vacc = sqrtf(cov_d_d)
```
I'll make the change to get in to testing. We can alter it later if needed

ericsenant: LGTM then

PR #499, opened 2020-09-10
UBX-NAV-SAT Support [TES-16]

benjaminaltieri:
Add handling of `UBX-NAV-SAT` message to generate `SBP_MSG_SV_AZ_EL` and `SBP_MSG_MEASUREMENT_STATE` messages used in NMEA GSV generation


c/src/gnss_converters/ubx_sbp.c
```diff
@@ -975,6 +977,84 @@ static void handle_nav_velecef(struct ubx_sbp_state *state, u8 *inbuf) {
...
+  msg_measurement_state_t *msg_meas_state =
+      (msg_measurement_state_t *)meas_state_buf;
+
+  for (int i = 0; i < nav_sat.num_svs; i++) {
+    ubx_nav_sat_data *data = &nav_sat.data[i];
+
+    u8 sat_id = data->sv_id;
+    u8 code = convert_ubx_gnssid_sigid(data->gnss_id, data->sv_id);
+
+    msg_az_el->azel[i].sid.sat = sat_id;
```
woodfell, line 1005: I think this is going to overflow if a single NAV-SAT generates several SV_AZ_EL output messages. There needs to be another counter for the index into the SBP message which gets reset to 0 every time a message gets sent early.

c/src/gnss_converters/ubx_sbp.c
```diff
@@ -975,6 +977,84 @@ static void handle_nav_velecef(struct ubx_sbp_state *state, u8 *inbuf) {
...
+    msg_az_el->azel[i].sid.code = code;
+    msg_az_el->azel[i].az = lround((double)data->azim / 2.0);
+    msg_az_el->azel[i].el = data->elev;
+    azellen += sizeof(sv_az_el_t);
+
+    msg_meas_state->states[i].mesid.sat = sat_id;
```
woodfell, line 1011: Same here, an extra counter required to replace `i` in `msg_meas_state->states[i]`

c/src/gnss_converters/ubx_sbp.c
benjaminaltieri: Good catch, will fix that up

c/src/gnss_converters/ubx_sbp.c
benjaminaltieri: should be fixed

woodfell (approved):
LGTM!

anth-cole (approved):
lgtm - I think this should enable the GSV out message

c/tests/check_ubx.c
```diff
@@ -206,6 +207,47 @@ static void ubx_sbp_callback_rxm_sfrbx_sbas(
...
+  if (msg_id == SBP_MSG_MEASUREMENT_STATE) {
+    msg_measurement_state_t *msg = (msg_measurement_state_t *)buff;
+    ck_assert_uint_eq(msg->states[0].mesid.sat, 0);
+    ck_assert_uint_eq(msg->states[0].mesid.code, CODE_GPS_L1CA);
+    ck_assert_uint_eq(msg->states[0].cn0, 12);
+    ck_assert_uint_eq(msg->states[1].mesid.sat, 1);
+    ck_assert_uint_eq(msg->states[1].mesid.code, CODE_GAL_E1B);
+    ck_assert_uint_eq(msg->states[1].cn0, 16);
+    ck_assert_uint_eq(msg->states[2].mesid.sat, 2);
+    ck_assert_uint_eq(msg->states[2].mesid.code, CODE_BDS2_B2);
```
ericsenant, line 245: Isn't that the Code for B2I? To me the code for B1I is CODE_BDS2_B1

benjaminaltieri: There may be an error in the translate coding, but this is where I took those numbers from: https://github.com/swift-nav/gnss-converters/blob/6c6d3fbf88b6937a13f63ff146971e1ec939973f/c/src/gnss_converters/ubx_sbp.c#L255-L258

ericsenant: The codes reference is in https://github.com/swift-nav/libswiftnav/blob/master/include/swiftnav/signal.h#L236

benjaminaltieri: The code I linked to is where the conversion between ubx sat id and our code is taking place. It is called from here: https://github.com/swift-nav/gnss-converters/blob/6c6d3fbf88b6937a13f63ff146971e1ec939973f/c/src/gnss_converters/ubx_sbp.c#L1002

benjaminaltieri: I am not sure if `convert_ubx_gnssid_sigid` is correct or not, as I am re-using it without inspection. I'm hoping you can help identify if there is an error there 😄

ericsenant: I looked at that one, but I was not sure what was decoded from ubx at first, and could not find ubx/decode_ubx.c : where is that?

mfine: @ericsenant it's in here: https://github.com/swift-nav/libubx/blob/master/c/src/decode.c

c/src/gnss_converters/ubx_sbp.c
```diff
@@ -975,6 +976,84 @@ static void handle_nav_velecef(struct ubx_sbp_state *state, u8 *inbuf) {
...
+  memset(meas_state_buf, 0, SBP_MAX_PAYLOAD_LEN);
+  u8 meas_cnt = 0;
+  msg_measurement_state_t *msg_meas_state =
+      (msg_measurement_state_t *)meas_state_buf;
+
+  for (int i = 0; i < nav_sat.num_svs; i++) {
+    ubx_nav_sat_data *data = &nav_sat.data[i];
+
+    u8 sat_id = data->sv_id;
+    u8 code = convert_ubx_gnssid_sigid(data->gnss_id, data->sv_id);
```
ericsenant, line 1002: here I believe this is not an sv_id required as second parameter, but a sig_id defining the type of signal. 
With the ublox M8L, you have sig_id=0 for all GNSS_IDs; Not sure how to make this generic for handling F9 as well.

benjaminaltieri: Ok I think I finally understand now.. we don't get the signal id from the NAV SAT message, so we can't know which signal each entry corresponds to. Well I will hardcode to zero and I can place a comment on the code and hopefully we will understand how to handle F9P

benjaminaltieri: https://github.com/swift-nav/gnss-converters/pull/512

PR #503, opened 2020-09-10
[ubx2sbp]: Output SBP_MSG_TIME_GPS messages

dgburr:
Output SBP_MSG_TIME_GPS messages before SBP_MSG_POS_LLH messages since some tools (e.g. sbp2report) use SBP_MSG_TIME_GPS to perform time range filtering


c/src/gnss_converters/ubx_sbp.c
```diff
@@ -525,6 +525,18 @@ static int fill_msg_pos_llh_hnr(const u8 buf[], msg_pos_llh_t *msg) {
   return 0;
 }
 
+static int fill_msg_gps_time(struct ubx_sbp_state *state, msg_gps_time_t *msg) {
+  if (!state->obs_state.weeknumber_set) {
+    return -1;
+  }
+
+  msg->wn = state->obs_state.last_obs_time_gnss.wn;
+  msg->tow = state->last_tow_ms;
```
reimerix, line 534: I think technically, the GNSS obs timestamp is not corrected for the receiver clock offset on u-blox receivers, whereas timestamps in e.g. `UBX-NAV-PVT` would be. There's a chance those two would be different by a couple of milliseconds and thus also the solution messages could carry a timestamp slightly different to the `MSG_GPS_TIME` timestamp.

reimerix (commented):
Code changes look good to me, but I think there might be better sources for the TOW field. Also, it would be good to add a test for the new conversion in `c/tests/check_ubx.c`.

PR #512, opened 2020-09-14
Fix gnss id to code conversion in NAV SAT conversion [TES-16]

benjaminaltieri:
The previous PR mistakenly used sat id instead of signal id to get the code for NAV SAT conversion, this fixes the value to 0

I have also added the SBAS code as I noticed those were being set to invalid.


ericsenant (approved):
lgtm

PR #524, opened 2020-09-18
Support fused wagon in sbp2nmea

woodfell:


c/src/gnss_converters/sbp_nmea.c
```diff
@@ -284,7 +324,23 @@ void sbp2nmea_to_str(const sbp2nmea_t *state, char *sentence) {
   state->cb_sbp_to_nmea(sentence, state->ctx);
 }
 
-void *sbp2nmea_msg_get(const sbp2nmea_t *state, sbp2nmea_sbp_id_t id) {
+void *sbp2nmea_msg_get(const sbp2nmea_t *state,
```
woodfell, line 327: The extra parameter here is to prevent duplication of code in all the `send_*` functions. When `consider_mode` is set to `true` this function will for any message where there is the possibility of a `_GNSS` decorate return the appropriate message for the currently selected mode. For example when in gnss mode calling `sbp2nmea_msg_get(state, SBP2NMEA_UTC_TIME)` will actually return `SBP2NMEA_UTC_TIME_GNSS`. Passing `false` in this parameter returns exactly what was requested

anth-cole (approved):
lgtm

c/src/gnss_converters/sbp_nmea.c
```diff
@@ -212,14 +243,54 @@ static bool sbp2nmea_discard_sbp(const sbp2nmea_sbp_id_t sbp_id,
   return false;
 }
 
+static sbp2nmea_sbp_id_t recode_legacy_sbp_id(const sbp2nmea_t *state,
```
woodfell, line 246: I admit this isn't great, it's here to support older log files before the `_GNSS` decorate was available, it also makes the unit tests in this repo pass. Short term fix, doesn't actually come in to play for the actual usage of this rework in starling

c/src/gnss_converters/sbp_nmea.c
```diff
@@ -212,14 +243,54 @@ static bool sbp2nmea_discard_sbp(const sbp2nmea_sbp_id_t sbp_id,
...
+
+static void update_mode(sbp2nmea_t *state, sbp2nmea_sbp_id_t sbp_id) {
+  if (state->requested_mode == SBP2NMEA_MODE_BEST &&
+      state->actual_mode == SBP2NMEA_MODE_GNSS &&
+      sbp_id == SBP2NMEA_SBP_GROUP_META) {
+    state->actual_mode = SBP2NMEA_MODE_FUSED;
+  }
+}
+
+static void check_fused_wagon_complete(sbp2nmea_t *state,
```
woodfell, line 270: This has returned to the old version of just looking for GROUP_META and SOLN_META to mark fused wagon completion. Doesn't actually count the messages sent in the wagon to avoid changing the sbp2nmea API (would have to replace sbp2nmea_sbp_id_t with u16 in sbp2nmea()), can be done in the future.

c/src/gnss_converters/nmea.c
```diff
@@ -1224,10 +1248,16 @@ void send_pubx(const sbp2nmea_t *state) {
...
-    NMEA_SENTENCE_PRINTF("%.1f,%.1f,%.1f,",
-                         round(10 * sbp_dops->hdop * 0.01) / 10,
-                         round(10 * sbp_dops->vdop * 0.01) / 10,
-                         round(10 * sbp_dops->tdop * 0.01) / 10);
+    if (state->actual_mode == SBP2NMEA_MODE_GNSS) {
+      NMEA_SENTENCE_PRINTF("%.1f,%.1f,%.1f,",
+                           round(10 * sbp_dops->hdop * 0.01) / 10,
+                           round(10 * sbp_dops->vdop * 0.01) / 10,
+                           round(10 * sbp_dops->tdop * 0.01) / 10);
+    } else {
```
woodfell, line 1256: tdop isn't available in fused wagon

benjaminaltieri (approved):
LGTM!

PR #530, opened 2020-09-23

dgburr:
The current implementation of pack_carrier_phase() and pack_doppler() incorrectly rounds values which are very close to a whole number. For example, 1613.994385 should be represented as 1613/255, not 1614/0.


dgburr:
This PR transfers to https://github.com/swift-nav/gnss-converters/pull/413 to starling-v1.2.0-release and includes the additional fix from https://github.com/swift-nav/gnss-converters/pull/406/files to prevent e.valid from being overwritten

pgrgich (approved):
Lgtm.

PR #540, opened 2020-09-30
Do not convert eph rtcm mess to sbp if GLO sat = 0 (#539)

sasamb:
With this PR, we prevent the GLO 0 ephemeris RTCM message to be translated in SBP. 
When Starling processes an ephemeris for sat=0, it crashes. See the [bug report on Slack](https://snav.slack.com/archives/CDBCP43BM/p1601021501005200).
We have already fixed Skylark to not send out GLO 0 ephemeris, but it would be better if Starling had a kind of protection against this issue. cc @scarcanague 
This change has already been merged in Gnss-converters master.


benjaminaltieri (approved):
IGTM, let's submit for 1.2.20 hot fix as well

PR #543, opened 2020-10-01
Change hardcoded GPS ephemeris P code fields to the values always sen…

ljbade:
…t by the sats


c/src/gnss_converters/sbp_rtcm3_ephemeris.c
```diff
@@ -35,8 +35,8 @@ void sbp_to_rtcm3_gps_eph(const msg_ephemeris_gps_t *sbp_gps_eph,
       rtcm3_encode_fit_interval_gps(sbp_gps_eph->common.fit_interval);
 
   // SBP doesn't include this information so it is hardcoded
-  msg_eph->kepler.codeL2 = 3;         /* L2C enabled */
-  msg_eph->kepler.L2_data_bit = true; /* L2P Nav data off */
+  msg_eph->kepler.codeL2 = 1;         /* P-code enabled */
```
scarcanague, line 38: It seems the spec is ambiguous, because L2C is also on right so it could be 3?

scarcanague (approved):
One question but makes sense

c/src/gnss_converters/sbp_rtcm3_ephemeris.c
ljbade: image

I believe the RTCM spec is wrong (or out of date), current GPS ICD (above) says value 3 is invalid and that this field only talks about the in-phase component, not the L2C phase.

PR #570, opened 2020-10-28

pgrgich:
…ly when it has both.


lloydmaza (approved):
LGTM. 

Any reason we suppress the invalid-flagged messages? Why not convert them anyway?

dgburr:
If we want to keep the behaviour consistent with ubx2sbp then the behaviour should be to convert observations even if the pseudorange or carrier phase is invalid: https://github.com/swift-nav/gnss-converters/blob/master/c/src/gnss_converters/ubx_sbp.c#L386

pgrgich:
@lloydmaza @dgburr I was wondering what to do, but went with this initially because it will reduce throughput with no downside (clearly signals with invalid code and phase can never be used). The counter-argument is that the converter should be dumb, so it's easy to know if something went wrong in it (num obs in must equal num obs out). I wasn't aware of the ubx not doing this, so we should make the two consistent, regardless of which way we choose.

What is your preference? @scarcanague how about you?

scarcanague:
I see, I would be in favor of using the `Invalid pseudorange measurement` and `Invalid carrier phase measurement` of the SBP instead of dropping, as I feel like a year from now when we'll compare ST internal PE and our SBP obs stream and we see a discrepancy, that may trigger a new unnecessary investigation

PR #571, opened 2020-10-28

pgrgich:
…ly when it has both.


pgrgich:
Closed in favor of https://github.com/swift-nav/gnss-converters/pull/570.

PR #572, opened 2020-10-28

pgrgich:
…ly when it has both.


c/tests/check_rtcm3.c
```diff
@@ -99,22 +99,6 @@ static packed_obs_content_t sbp_test_data[] = {
     {1145898151, {93845224, 62}, {-539, 207}, 148, 14, 15, {15, 1}},
     {1022557545, {83744032, 221}, {-215, 37}, 188, 14, 15, {27, 1}},
     {1222990548, {100158785, 33}, {-704, 56}, 132, 13, 15, {30, 1}},
```
pgrgich, line 101: These code phase values are consistent for L1 (i.e. code-phase on L1 is consistently ~2500m) using the frequency numbers I put in here: https://github.com/swift-nav/gnss-converters/pull/572/files#diff-8889aa2f17aafc4040c39eabe1b7dea1d2a05df8e9e8c4b15ec0e7a6aec61102R1494-R1501, but then the L2 code-phase differences fluctuate from -16000 - +12000m. On the other hand, if I set the frequency numbers to zero for all satellites, the L1 code-phase difference is within meters of the L2 code-phase difference for each satellite, but the differences fluctuate from -55000 - +44000m. This leads me to believe that the L1 values were taken from real data and the L2 derived on the assumption of zero frequency numbers, or some such.

In any case, neither option works, because RTCM requires the code phase difference for all signal/satellite pairs to be small, otherwise the phase is invalidated. Is it worthwhile correcting these values for the test or shall we simply delete them? They are only used in the RTCM round trip test modified below.

pgrgich:
Closed in favor of https://github.com/swift-nav/gnss-converters/pull/570.

